PR #388, opened 2024-03-06
[-] fix possible index out of bound issue, fixes #387

### LINKED ISSUE #387: panic at `pgwatch3/log/log_broker_hook.go`
**Describe the bug**
pgwatch3 die after a few minutes of idle
 

**Steps to produce the behavior:**
1. Follow the readme to add test db and bench
2. poke around the webui and wait for a while?
3. panic

~~I haven't been able to reproduce it though~~ reproduced (but the situation is kinda like iterator's use after invalidation)

**Screenshots**
~~I wonder if its also related to the sql error above~~
Not related but weird
image

```
panic: runtime error: index out of range [1] with length 1

goroutine 902 [running]:
github.com/cybertec-postgresql/pgwatch3/log.(*BrokerHook).RemoveSubscriber(0xc0000a7110, 0xc000618420)
	/pgwatch3/log/log_broker_hook.go:59 +0x205
github.com/cybertec-postgresql/pgwatch3/webserver.writer(0xc0000b3080, {0x1638860, 0xc00003ebd0})
	/pgwatch3/webserver/wslog.go:64 +0x2a2
created by github.com/cybertec-postgresql/pgwatch3/webserver.(*WebUIServer).serveWsLog in goroutine 915
	/pgwatch3/webserver/wslog.go:82 +0x1ba
```

Fix: iterate from back to front so you don't have to worry about indexes that are deleted

qazxcdswe123:
Close #387

by doing it backward it should be fine

### DIFF
--- a/src/log/log_broker_hook.go
+++ b/src/log/log_broker_hook.go
@@ -2,6 +2,7 @@ package log
 
 import (
 	"context"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pashagolub:
Thanks for your help! I'll check this ASAP!

pashagolub:
How do you feel about such a fix? Does it work for you?

qazxcdswe123:
Hi, this fix is definitely better than my approach. I've also added a test and can confirm that the previous buggy implementation is now fixed

pashagolub (approved):

PR #394, opened 2024-03-12
Added pwctl support with check pgwatch3 status (Partial)

pranayjoshi:
This Pull request adds the pwctl support to pgWatch3.

Issues with this pull:

This pull is not yet integrated with the `main()` function. i.e it actually won't work. But the code is locally tested and works as expected. **I am waiting for the maintainers response** to make it fully functional and add the required lines to main() function and to the dockerfile.

Also, please guide me to the contributers guide, if any.

Any form of review and critiques is really appreciated.
Thank you

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,0 +1,29 @@
+# Makefile
+
+# The name of the Go binary to build
+BINARY_NAME=pwctl
+
+# The directory to install the binary to
+INSTALL_DIR=/usr/local/bin
+
+# The Go command
+GO=go
+
+# The default target of this Makefile
+all: permissions
+
+# Compiles the Go program
+build:
+	$(GO) build -o $(BINARY_NAME)
+
+# Installs the Go program as a command-line tool
+install: build
+	mv $(BINARY_NAME) $(INSTALL_DIR)
+
+# Provide necessary permissions to the binary
+permissions: install
+	chmod +x $(INSTALL_DIR)/$(BINARY_NAME)
+
+# Cleans up the binary
+clean:
+	if [ -f $(INSTALL_DIR)/$(BINARY_NAME) ] ; then rm $(INSTALL_DIR)/$(BINARY_NAME) ; fi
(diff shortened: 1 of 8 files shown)


src/Makefile
pashagolub, line 1: I don't want makefiles in this projet. We use Go-specific ways, e.g. goreleaser

pashagolub (commented):
I'm not sure if we should use `cobra` for such a simple project. I'm afraid it will add extra dependencies. Have you considered other minimalistic ways?

As for folder structure, since pwctl supposed to be standalone binary, this looks promising
```
- cmd
  - pwctl
    - main.go
      *.go
      go.mod
      go.sum
```

muni106:
be careful that your isPortActive function only checks if the port is active, while I think the goal of the "status" command is to check that pgwatch is running on the port

pranayjoshi:
@muni106 Hi please review the full code... IsPortActive is 1 of the 3 functions. Its not even the main function

pranayjoshi:
> I'm not sure if we should use `cobra` for such a simple project. I'm afraid it will add extra dependencies. Have you considered other minimalistic ways?
> 
> As for folder structure, since pwctl supposed to be standalone binary, this looks promising
> 
> ```
> - cmd
>   - pwctl
>     - main.go
>       *.go
>       go.mod
>       go.sum
> ```

@pashagolub  That would make my life so much easier 😄 as now I dont have to deal with clashing with other dependency. Thanks

pashagolub:
No problem! No, of course we can use the existing library. :-) I think there are some lighter. But if you think cobra is the way to go, put it into proposal. I'll gladly consider it as a way to go.

pranayjoshi:
Hey @pashagolub , I researched on some light-weight cli libraries and found `urfave/cli` to be the best in our case. Shall go forward with it?

pashagolub:
It's your task to choose how to proceed. :-) I have no preferences.

PR #465, opened 2024-06-17
[!] add RPC sink support, closes #464

### LINKED ISSUE #464: RPC sinks for Pgwatch3
**Feature**: To add support for sinks on remote nodes.

**Description**: A simple RPC based interface to interact with remote sinks where pgwatch3 can send the measurements and the sink processes the data according to its configuration/type.

destrex271:
Addresses Issue #464 .

Still WIP.


src/sinks/types.go
```diff
@@ -0,0 +1,12 @@
+package sinks
+
+import (
+	"github.com/cybertec-postgresql/pgwatch3/metrics"
+)
+
+type MeasurementMessage metrics.MeasurementMessage
+
+type WriteRequest struct {
```
pashagolub: Meh. I don't like this. We have `metrics.MeasurementMessage.CustomTags` for that. Every sink writer can update it with whatever it needs to, e.g.

```go
msg.CustomTags['pgwatchID'] = 2
```

src/sinks/rpc.go
```diff
@@ -0,0 +1,76 @@
+/*
+*
+* RPC sink implementation for pgwatch3.
+* Requires the address and port of the sink.
+*
+ */
+
```
pashagolub:
```suggestion
```

All files must follow the same format. There are no headers in other files, thus we don't want any header here too.

src/sinks/rpc.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	for _, msg := range msgs {
+		var status int
+		pgwatchID := os.Getenv("pgwatch_id")
+		msg.CustomTags = make(map[string]string)
+		if len(pgwatchID) > 0 {
+			msg.CustomTags["pgwatchId"] = pgwatchID
+		} else {
+			msg.CustomTags["pgwatchId"] = strconv.Itoa(os.Getpid()) + "_pgwatch3" // Replaces with PID to create a pgwatchid
+		}
```
pashagolub: We don't want any `pgwatchId`. If we want to distinguish pgwatch instances in sinks, we can use `MeasurementMessage.SystemIdentifier` for that purpose.

As I told, `os.Getpid()` will return 1 for all Docker containers which gives us nothing. But even getting proper pid on a system gives us nothing in terms of debug. 

Speaking of this I want such a declaration on a receiver side:

```go
func (receiver *RPCWriter) UpdateMeasurements(msgs []metrics.MeasurementMessage, status *int) error 
```

Probably receiver implementation needs to import `"github.com/cybertec-postgresql/pgwatch3/metrics"` to use declarations in place, instead of re-declaring it locally.

src/sinks/rpc.go
```diff
@@ -0,0 +1,76 @@
...
+		err := rw.client.Call("Receiver.UpdateMeasurements", &msg, &status)
```
pashagolub:
```suggestion
		err := rw.client.Call("RPCWriter.UpdateMeasurements", &msg, &status)
```

according to the previous comment

src/sinks/rpc.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return nil
+}
+
+func (rw *RPCWriter) SyncMetric(_, _, _ string) error {
+	if rw.ctx.Err() != nil {
+		return rw.ctx.Err()
+	}
+	return nil
+}
```
pashagolub: This one should be implemented calling appropriate `RPCWriter` method. It's omited for `JSONWriter` but it should be implemented. For example, RPC `CSVWriter` might want to open new file when a new database and/or metric added. Or it might want to close file and zip it, of `op=="remove"`

src/sinks/rpc_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"net"
+	"net/http"
+	"net/rpc"
+	"testing"
+
+	"github.com/cybertec-postgresql/pgwatch3/metrics"
+)
+
+type Receiver struct {
```
pashagolub, line 17: `Receiver` should implement and test `SyncMetric()` as well

src/sinks/rpc.go
```diff
@@ -0,0 +1,100 @@
...
+		if len(pgwatchID) > 0 {
+			msg.CustomTags["pgwatchId"] = pgwatchID
+		} else {
+			msg.CustomTags["pgwatchId"] = strconv.Itoa(os.Getpid()) + "_pgwatch3" // Replaces with PID to create a pgwatchid
+		}
+		err := rw.client.Call("Receiver.UpdateMeasurements", &msg, &status)
+		if err != nil {
+			return err
+		}
+        rw.SyncMetric("demo", "abc", "ADD")
```
pashagolub: it shouldn't be here. test leftover?

pashagolub:
Please, resolve all requested changes

pashagolub (approved):

PR #473, opened 2024-07-05
[*] update monitored sources grid, closes #445

### LINKED ISSUE #445: update monitored sources grid

denys-holub:
### Summary
- Updated sources grid page
- Updated create, update and copy source form


pashagolub:
- [ ] change `PATCH` for metrics to `POST` 
- [ ]

denys-holub:
Found a bug for UI, after adding/updating something, try to open again form for adding/updating. UI will fail.
Screenshot 2024-07-08 at 09 45 47

denys-holub:
@pashagolub 
Also found a bug for update, after submiting updating form, element is not updated. But appears new element with updated props

denys-holub:
Updated all `PATCH` methods for `POST`, but create is working instead of update.

On `POST /source` there's no `name` query parameter
Screenshot 2024-07-08 at 10 59 41

For metrics and presets endpoints expect for `name` query parameter, but for update we should somehow send two names: `oldName & newName`, correct me if I'm wrong

pashagolub:
I suggest to send `delete` and `insert` requests if the name changes.

pashagolub:
I suggest to send `delete` and `insert` requests if the name changes.

pashagolub (approved):

PR #527, opened 2024-08-27
[-] fix webui if opened from a remote host, fixes #526

### LINKED ISSUE #526: WebUI is not working from remote host
### Discussed in https://github.com/cybertec-postgresql/pgwatch/discussions/523

<div type='discussions-op-text'>

<sup>Originally posted by **ondar** August 26, 2024</sup>

When I log in from another computer on the network, I get Undefined error after pressing login:
image
Seems UI is requesting response from localhost and not the remote server I run it from.

Backend and Grafana seems to be working just fine. What am I doing wrong?</div>

pashagolub:


internal/webui/src/axiosInstance.ts
```diff
@@ -1,18 +0,0 @@
-import axios, { AxiosError } from "axios";
-import { getToken } from "services/Token";
-
-export const axiosInstance = axios.create({
-  baseURL: "http://localhost:8080/"
```
ondar, line 5: So I was correct after all 😁

ondar (approved):
Thanks, looks good.

internal/webui/src/axiosInstance.ts
pashagolub: Yup, thanks for pointing this out!

PR #536, opened 2024-08-29
refactor: Updated Receiver to Handler for Sync Metric functionality. …

destrex271:
References the current upstream version of pgwatch3_rpc_receivers where we have separated both the receiver and the sync metric handler.

A new type called Handler now handles the Sync Metric calls in the receivers: https://github.com/destrex271/pgwatch3_rpc_server/blob/1cb1f44d237ce5817831194c9285c20397277fc2/cmd/csv_receiver/main.go#L32

### DIFF
--- a/internal/sinks/rpc.go
+++ b/internal/sinks/rpc.go
@@ -64,7 +64,7 @@ type SyncReq struct {
 
 func (rw *RPCWriter) SyncMetric(dbUnique string, metricName string, op string) error {
 	var logMsg string
-	if err := rw.client.Call("Receiver.SyncMetric", &SyncReq{
+	if err := rw.client.Call("Handler.SyncMetric", &SyncReq{
 		Operation:  op,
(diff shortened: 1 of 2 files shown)


pashagolub:
No, we will keep it `Receiver`. I don't see any point in renaming things. It is the same object handling all functions, at the moment `UpdateMeasurements` and `SyncMetrics`. If RPC sink has two different objects dealing with it, it's the choice of the author. pgwatch considers RPC sink as a single interface supporting all (at the moment 2) methods.

destrex271:
> No, we will keep it `Receiver`. I don't see any point in renaming things. It is the same object handling all functions, at the moment `UpdateMeasurements` and `SyncMetrics`. If RPC sink has two different objects dealing with it, it's the choice of the author. pgwatch considers RPC sink as a single interface supporting all (at the moment 2) methods.

Ok! I'll make the required changes in receivers repo

PR #546, opened 2024-09-19
[-] fix Patroni source resolving

saletnik:
Fixed broken patron support

### DIFF
--- a/internal/sources/resolver.go
+++ b/internal/sources/resolver.go
@@ -16,13 +16,12 @@ import (
 	"os"
 	"path"
 	"regexp"
-	"strconv"
+	"strings"
 	"time"
 
 	"github.com/cybertec-postgresql/pgwatch/v3/internal/db"
 	"github.com/cybertec-postgresql/pgwatch/v3/internal/log"
 	pgx "github.com/jackc/pgx/v5"
-	"github.com/jackc/pgx/v5/pgxpool"
 	client "go.etcd.io/etcd/client/v3"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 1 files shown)


internal/sources/resolver.go
```diff
@@ -173,14 +173,24 @@ func getEtcdClusterMembers(s Source) ([]PatroniClusterMember, error) {
...
+		// Key="/service/batman/leader"
+		// Key="/service/batman/members/node"
+		//
+		// create unique map of first level items without Namespace
+		scopes := make(map[string]bool, len(resp.Kvs))
 		for _, node := range resp.Kvs {
-			scope := path.Base(string(node.Key)) // Key="/service/batman"
+			pathSuffix := strings.TrimPrefix(string(node.Key), s.HostConfig.Namespace)
+			scope := strings.SplitN(pathSuffix, "/", 2)[0]
+			scopes[scope] = true
```
pashagolub, line 190: Why do we need this?

saletnik: etcd returns keys: 
```
/service/scope1/config 
/service/scope1/initialize 
/service/scope1/leader 
/service/scope1/members/node1 
/service/scope1/members/node2 
/service/scope1/status 
/service/scope1/sync
/service/scope2/config 
/service/scope2/initialize 
/service/scope2/leader 
/service/scope2/members/node1 
/service/scope2/members/node2 
/service/scope2/status 
/service/scope2/sync
```

keys do not exist:
```
/service/scope1
/service/scope2
```

so I extract the fragment from the key
```
scope1
scope2
```

pashagolub: but what `scopes[scope] = true` means? And where the check for it? And why not False?

saletnik: After extracting the fragment we are left with 7 scope1  elements and 7 scope2 elements. To get rid of duplicates we add them to maps, the value in the map does not matter. Keys in the map guarantee uniqueness

pashagolub: ok, I got it. You're emulating set with maps. Thanks

PR #570, opened 2024-10-23
[*] fix typos and old info in docs and comments, fix non-ascii char in postgres.go, fix legacy backticks in build.sh

ondar:
I recommend going through docs again to spot and fix outdated info from pgwatch2. I fixed what I knew.


docs/concept/components.md
```diff
@@ -16,19 +16,19 @@ from.
 The metrics collector, written in Go, is the only mandatory and most
 critical component of the whole solution. The main task of the pgwatch
 collector / daemon is pretty simple - reading the configuration and
-metric defintions, fetching the metrics from the configured databases
-using the configured connection info and finally storing the metrics to
+metric definitions, fetching the metrics from the configured databases
+using the configured connection info and finally storing the metric results to
```
pashagolub:
```suggestion
using the configured connection info and finally storing the metric measurements to
```

internal/build.sh
```diff
@@ -3,7 +3,7 @@
...
 set -x
 
-GIT_HASH=`git show -s --format=%H HEAD`
-GIT_TIME=`git show -s --format=%cI HEAD`
-VERSION=`git rev-parse --abbrev-ref HEAD`
-go build -ldflags "-X 'main.commit=$GIT_HASH' -X 'main.date=$GIT_TIME' -X 'main.version=$VERSION'"
\ No newline at end of file
+GIT_HASH=$(git show -s --format=%H HEAD)
+GIT_TIME=$(git show -s --format=%cI HEAD)
+VERSION=$(git rev-parse --abbrev-ref HEAD)
```
pashagolub, line 8: Why do you prefer this format over backticks?

ondar: I don't really care, but Idea recommended this change. Seems backticks are legacy now (=outdated?) now? Stack Overflow is also suggesting that: https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts

pashagolub: ok, seems reasonable :)

pashagolub (approved):

pashagolub:
Thanks!

ondar:
@pashagolub welcome 😁 It seems I'm now pgwatch ambasador in our company so its my interest too that its the best we can make it.

PR #578, opened 2024-10-29
[*] improve Prometheus sink and Docker setup docs

ondar:


internal/cmdopts/cmdoptions.go
```diff
@@ -181,7 +181,7 @@ func (c *Options) NeedsSchemaUpgrade() (upgrade bool, err error) {
 
 // InitWebUI initializes the web UI server
 func (c *Options) InitWebUI(fs fs.FS, logger log.LoggerIface) error {
-	if c.WebUI.WebAddr == "" {
+	if c.WebUI.WebAddr == "" || c.WebUI.WebAddr == "''" {
```
pashagolub: It's getting uglier and uglier :-D 

Probably I should add `--web-disable` command-line option

ondar: It was the most my meager GO skills permitted 😁 Feel free to make it better.

pashagolub: It's not about you, it's about my laziness. :) I thought we can mimic behavior without adding new cmdopt. I was wrong, #579 is here to fix

ondar: Better 😁 Let me integrate it to this PR.

pashagolub: Already did

docs/tutorial/docker_installation.md
```diff
@@ -190,3 +190,40 @@ collector could be organized. For another example how various components
...
 [here](https://github.com/cybertec-postgresql/pgwatch/blob/master/docker-compose.yml).
+
+## Example of advanced setup using YAML files and dual sinks:
+
+pgwatch service in file `docker/docker-compose.yml` can look like this:
+```yaml
+  pgwatch:
+    image: cybertecpostgresql/pgwatch:latest
+    command:
+      - "--web-addr=''"
```
ondar: ah, you changing it, ok 😁 so change it in this doc file too (2 places - here and in the end)

pashagolub: On it

docs/tutorial/docker_installation.md
```diff
@@ -190,3 +190,37 @@ collector could be organized. For another example how various components
...
+  conn_str: postgresql://pgwatch:pgwatchadmin@postgres/pgwatch'
+  preset_metrics: exhaustive
+  is_enabled: true
+  group: default
+```
+
+Running this setup you get pgwatch that uses sources from YAML file and
+outputs measurements to postgres DB and exposes them for Prometheus 
+to scrape on port 8080 instead of WebUI (which is disabled by `--web-disable`).
+Metrics definition are built-in, you can examine definition in [`internal/metrics/metrics.yaml`](https://github.com/cybertec-postgresql/pgwatch/blob/master/internal/metrics/metrics.yaml).
```
ondar, line 233: This is true, but I wanted to show here that you can provide your own metric definition file.

pashagolub: Then we need to make one :) Otherwise people will just copy-paste and I want they know that metrics are always built-in

pashagolub: And it's a bad style to link to `internal` folder in Go projects anyway

ondar: All true, not only you are the lazy one 😁

pashagolub: How about to add

Metric definition file `metrics.yaml` in the same directory will define only one metric `db_size` for demonstration purposes. In real life you will have tens metrics.
```yaml
metrics:
    db_size:
        sqls:
            11: |-
                select /* pgwatch_generated */
                  pg_database_size(current_database()) as size_b,
                  (select sum(pg_total_relation_size(c.oid))::int8
                   from pg_class c join pg_namespace n on n.oid = c.relnamespace
                   where nspname = 'pg_catalog' and relkind = 'r'
                  ) as catalog_size_b
        gauges:
            - '*'
        statement_timeout_seconds: 300
```

pashagolub: But we can skip this part now. I will add a special how-to chapter about custom metrics anyway

ondar: Up to you 😁 You are the owner in the end, no?

pashagolub:
Thanks for the contribution! 🙇

Issue #573, opened 2024-10-28
Hardcoded users in initialization scripts


krisavi:
Why not use role that is assigned from connection string. In my environment I do not have such role and I do not wish to start creating new ones. 

https://github.com/cybertec-postgresql/pgwatch/blob/c779f60a65cec35b9adc6ad5683bfbf5a4692916/internal/metrics/postgres_schema.sql#L2

krisavi:
It is not that simple, the line 1 also includes role name. Most likely AUTHORIZATION 'pgwatch'; is not used as when you already are logged in with that user, then it does it for you. Will try to get it working for myself before making any pull request.

pashagolub:
The idea behind this is this script might be executed by superuser but we want all objects belong to monitor role

krisavi:
That script at the same time is executed by pgwatch itself. If that should be run as superuser, then I do not think having application connecting as superuser is a way to go. 
Superuser script that creates prepares user for pgwatch should be run outside of application from separate script.

pashagolub:
Yes, that's true now. I meant v2

pashagolub:
I've rechecked the code. This sql should be executed during `pgwatch config init` command which supposed to be run by the user with `CREATE` privileges. I will update the docs

pashagolub:
Fixed by #580

PR #580, opened 2024-10-31
Remove hardcoded pgwatch user in config schema creation

krisavi:
#573


pashagolub:
please do not create pull requests from your `master` branch. I cannot push my chages

PR #615, opened 2025-01-09
[+] add `/readiness` and `/liveness` endpoints, closes #611

### LINKED ISSUE #611: Add Liveness and Readiness probes for deployment in K8S
No

We operate PGWatch in a Kubernetes cluster and need to add standard endpoints for liveness and readiness probes to PGWatch.

Thanks

pashagolub:


internal/webserver/webserver.go
```diff
@@ -49,7 +58,11 @@ func Init(opts CmdOpts, webuifs fs.FS, mrw metrics.ReaderWriter, srw sources.Rea
 	mux.Handle("/preset", NewEnsureAuth(s.handlePresets))
 	mux.Handle("/log", NewEnsureAuth(s.serveWsLog))
 	mux.HandleFunc("/login", s.handleLogin)
-	mux.HandleFunc("/", s.handleStatic)
+	mux.HandleFunc("/liveness", s.handleLiveness)
+	mux.HandleFunc("/readiness", s.handleReadiness)
+	if !opts.WebDisable {
+		mux.HandleFunc("/", s.handleStatic)
```
ondar, line 68: Maybe add more routes here when UI is disabled? Like all the rest endpoints for UI, like /login, /metric, ...

pashagolub: Well, what if one wants to disable web UI but be able to control everything with REST API calls?

pashagolub: Maybe it's better to add `--rest-api-disable`? What are your concerns?

ondar: Concerns are security and resource usage. I don't mind if you introduce new command line parametr or just disable it right away. Up to you.

pashagolub: Should work now as intended, e.g.
`pgwatch --web-disable` or `pgwatch --web-disable=all` will disable both REST API and web UI,
while `pgwatch --web-disable=ui` will serve only REST API

ondar: Hmmm... but what if you want to disable UI and REST, but keep liveness and readiness probes? Add "AllButK8S" option? Or treat web-disable as list and add k8s options so you can use `pgwatch --web-disable=ui,rest`?

pashagolub: I'd say either you want REST API, or you don't want REST API. Frankly speaking, I don't see the problem in enabled REST API at all

pashagolub: I cannot think of any program that will try to disable/enable part of REST API for any purposes. For example, patroni REST is always on.

pashagolub: If one wants to disable REST API, they can use other liveness probes, for example, try to pull something from Prometheus sink and see if it works or not

ondar: REST API is always a security problem. Liveness and readiness, that return only status code are OK. And different programs handle that differently so in the end its up to you. Or I might propose pull request if you want?

pashagolub: This is about a sane level of configuration. I don't think there is any need to end up with hundred options. All endpoints are protected with password at the moment, certificates are in todo list. If REST API is a security problem, one just disables it. Thanks for contributing. 👍

PR #631, opened 2025-01-31
[-] fix concatenated output of `pgwatch metric print-init`

agelwarg:
Add semi-colon to end of each statement so that concatenated SQL generated with "pgwatch metric print-init <preset>" is valid and can be piped to psql cli


pashagolub (approved):

pashagolub:
thanks!

PR #636, opened 2025-02-06
[-] fix checkpointer dashboards for PostgreSQL 17, fixes #613

### LINKED ISSUE #613: "Checkpoints" graph showing no data for Postgres 17
**Describe the bug**
The ***Checkpointer / Bgwriter / Block IO Stats*** dashboard will not show checkpoint stats on the ***Checkpoints*** graph for Postgres 17.

**To Reproduce**
1. Connect to a Postgres 17 instance.
2. Ensure the `checkpointer` metric is being collected.
3. View no data in the ***Checkpoints*** Grafana graph.

image

**Expected behavior**
To see checkpoint data in the ***Checkpoints*** graph.

**Desktop:**
 - OS: Linux
 - Browser Firefox
 - Version 133.0.3

**Additional context**
It looks like the `checkpointer` metric for Postgres 17 is placing checkpoint counts in the `num_timed` and `num_requested` fields, but the Grafana graph is looking for these values in the `bgwriter` table where they are stored for lower Postgres versions.

**For `bg_writer` metric**
image

**For `checkpointer` metric**
image

If we change the Grafana SQL to the following, we see proper checkpoint data for Postgres 17. I'm not sure how best to change things such that this graph does not break for lower Postgres versions still pulling from the `bgwriter` table.

```SQL
SELECT
  $__timeGroup(time, '$agg_interval'),
  sum(nullif(ct - ct_lag, 0)) as "num_timed",
  sum(nullif(cr - cr_lag, 0)) as "num_requested"
FROM (
  SELECT
    time,
    (data->'num_timed')::int8 as ct, lag((data->'num_timed')::int8) over w as ct_lag,
    (data->'num_requested')::int8 as cr, lag((data->'num_requested')::int8) over w as cr_lag
  FROM
    checkpointer 
  WHERE
    $__timeFilter(time)
    AND dbname = '$dbname'
  WINDOW w as (order by time)
) x
WHERE ct >= ct_lag and cr >= cr_lag
GROUP BY 1
ORDER BY 1
```

**After doing so, we see the data:**

image

ay-bh:
### Changes
- Adds support for PostgreSQL 17's pg_stat_checkpointer in the dashboard
- Modifies SQL query to union data from both bgwriter and checkpointer tables
- Maintains backward compatibility with pre-v17 PostgreSQL versions

closes #613

### DIFF
--- a/grafana/postgres/v10/checkpointer-bgwriter-stats.json
+++ b/grafana/postgres/v10/checkpointer-bgwriter-stats.json
@@ -86,7 +86,7 @@
           "orderByTime": "ASC",
           "policy": "default",
           "rawQuery": true,
-          "rawSql": "SELECT\n  $__timeGroup(time, '$agg_interval'),\n  sum(nullif(ct - ct_lag, 0)) as \"checkpoints_timed\",\n  sum(nullif(cr - cr_lag, 0)) as \"checkpoints_req\"\nFROM (\n  SELECT\n    time,\n    (data->'checkpoints_timed')::int8 as ct, lag((data->'checkpoints_timed')::int8) over w as ct_lag,\n    (data->'checkpoints_req')::int8 as cr, lag((data->'checkpoints_req')::int8) over w as cr_lag\n  FROM\n    bgwriter\n  WHERE\n    $__timeFilter(time)\n    AND dbname = '$dbname'\n  WINDOW w as (order by time)\n) x\nWHERE ct >= ct_lag and cr >= cr_lag\nGROUP BY 1\nORDER BY 1\n",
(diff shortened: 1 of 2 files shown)


ondar:
I'm not SQL expert, but don't we need `$__timeGroup` in 2nd select too?

ay-bh:
> I'm not SQL expert, but don't we need `$__timeGroup` in 2nd select too?

From what I understand, we usually keep the `$__timeGroup` in the outer query so all the rows from both sources end up in a single aligned time series. If we added `$__timeGroup` within the second sub-select as well, we’d risk each table being grouped on its own, potentially causing misaligned or maybe duplicated buckets. By using the union of the raw data and then grouping once, we ensure everything is aggregated consistently. That said, if you see a scenario where we really do need it in the second sub-select, I’d be interested to learn more.

ondar:
Seems my bad again 😄 I haven't seen there are 3 selects in there, not 2. It's kinda hard to read in escaped string format.

ondar (approved):

pashagolub:
Good job! Thanks. 

Would you please update "Checkpointer Write / Sync durations" panel as well, specifically `write_time` and `sync_time` from `checkpointer`.

PR #704, opened 2025-03-31
Add a SystemD service template to DEB / RPM / Tar

kmoppel:
To get started more quickly after installing from packages. Currently a bit annoying to hunt for a SystemD definition from old pgwatch2 and make source and sink adjustments


pashagolub:
Thanks!

PR #706, opened 2025-04-03
[*] deprecate `IsSuperuser` source option

pashagolub:
This will be determined during runtime


pashagolub:
Thanks @markwort for reporting!

PR #725, opened 2025-04-15
Update init_persistent_config.sh

iuuuuan:
Add newline for PW_GRAFANANOANONYMOUS


iuuuuan:
Patch is working as expected. File `/etc/grafana/grafana.ini` contain new line for block `[auth.anonymous]`

```
[server]
protocol = http
cert_file = /pgwatch/persistent-config/self-signed-ssl.pem
cert_key = /pgwatch/persistent-config/self-signed-ssl.key

[security]
admin_user = user
admin_password = password

[auth.anonymous]
enabled = true
# Organization name that should be used for unauthenticated users
org_name = Main Org.
# Role for unauthenticated users, other valid values are `Editor` and `Admin`
org_role = Editor

[dashboards]
default_home_dashboard_path = /var/lib/grafana/dashboards/1-global-db-overview.json

[metrics]
enabled = false

[plugins]
preinstall = marcusolsson-treemap-panel

[feature_toggles]
angularDeprecationUI = false
[auth.anonymous]
enabled = false
```

ondar:
I see it twice once with `enabled = true` in the middle of the file and then with `enabled = false` on the end?

iuuuuan:
> I see it twice once with `enabled = true` in the middle of the file and then with `enabled = false` on the end?

This is because template file [docker/demo/grafana.ini](https://github.com/cybertec-postgresql/pgwatch/blob/f573dbf78498daab28501920c1443b5819181f89/docker/demo/grafana.ini ) contains first `[auth.anonymous]` block. Second block is being added with [docker/bootstrap/init_persistent_config.sh](https://github.com/cybertec-postgresql/pgwatch/blob/f573dbf78498daab28501920c1443b5819181f89/docker/bootstrap/init_persistent_config.sh) script which overrides option `enabled`  for block `[auth.anonymous]`.

ondar:
If that is intended behavior then ok 😃

pashagolub:
I agree with @ondar. Having two entries in config looks weird. I'll prefer changing the value for the first entry, maybe with `sed`

iuuuuan:
Added new commit - change value with sed for auth.anonymous.

ondar:
Interesting regular. There is a space after `enabled = true` in the original file?

iuuuuan:
> Interesting regular. There is a space after `enabled = true` in the original file?

You mean [[:space:]]* which matches zero or more whitespace characters ? This is just in case if original file line with start with whitespace character but in current case you are right, it is redundant.

ondar:
Ah, true its * and not +, ok, thanks for clarification. Approving this PR.

ondar (approved):

PR #796, opened 2025-06-11
[+] add TLS encryption to RPC channel

0xgouda:
add new `root-ca $PW_RPC_ROOT_CA` flag for CA file path. 
use `tls` package to listen for connection and add user provided CA as a trusted authority. 
add rpc_tests_sinks dir containing test-only certificates to use in rpc_test.go


pashagolub:
Thanks for the PR.

I don't want to add additional option. I want all needed options for sinks to be present in the connection url.

For example, if one wants to use root CA with Postgres connection, they will just use [sslrootcert](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT) option in connection string, e.g.
```
pgwatch .. --sink=postgresql://user@host/dbname?sslrootcert=filename
```

I want to use the same approach for all sinks, so we don't bloat cmd options. This situation will hit us hard if one want to use two or more sinks but with different option values.

internal/sinks/rpc.go
```diff
@@ -2,23 +2,52 @@ package sinks
...
 
 	"github.com/cybertec-postgresql/pgwatch/v3/internal/log"
 	"github.com/cybertec-postgresql/pgwatch/v3/internal/metrics"
 )
 
-func NewRPCWriter(ctx context.Context, address string) (*RPCWriter, error) {
-	client, err := rpc.DialHTTP("tcp", address)
+func NewRPCWriter(ctx context.Context, path string) (*RPCWriter, error) {
+	address, params, found := strings.Cut(path, "?")
+	if (!found) {
```
pashagolub: What if I don't want to use TLS? Why do we force people?

internal/sinks/rpc.go
```diff
@@ -2,23 +2,52 @@ package sinks
...
+		return nil, fmt.Errorf("invalid RPC connection string, Usage: rpc://host?sslrootca=[CA_file]")
+	}
+
+	key, RootCA, found := strings.Cut(params, "=")
```
pashagolub: I'd like to use `net/url` for this, check https://pkg.go.dev/net/url#example-URL

internal/sinks/rpc.go
```diff
@@ -2,23 +2,52 @@ package sinks
...
+	if (!found || key != "sslrootca") {
```
pashagolub: We may just ignore unsupported parameters. What if server waits some additional params?

internal/sinks/rpc.go
0xgouda: updated, 
it now fallsback to the default `rpc.DialHTTP()` if `sslrootca` param was not provided
also deleted `address` param from the `RPCWriter` struct as it is never used
please check.

internal/sinks/rpc.go
0xgouda: done, 
but i think we may consider using it in the multiwriter itself instead of `strings.Cut()` and update sinks signature to receive relevant fields, i didn't do so now because it would have breaked the prometheus sink.

pashagolub: well, in multiwriter we're interested only on prefix so that's ok there. but when we try to work with params and values, it's easier to use built-in machinery

0xgouda: yeah i get its ok there, but i mean using built-in `net/url` will also do the job but will make use avoid calling it again when we want to work with params or other things inside one of the sinks

pashagolub:
Thanks a lot! Good job!

PR #849, opened 2025-07-17
[!] make recommendation metrics first-class citizens, closes #824

### LINKED ISSUE #824: all recommendations metrics are seen internally as one metric
**Describe the bug**
The `Reap()` method in reaper module, assigns all `reco_*` prefixed metrics the `recoMetricName` const name so if 2 or more `reco_*` metrics are present for certain database, they will be seen as the one `recoMetricName` metric and their cancel functions and metric definitions will interfere 
```
if strings.HasPrefix(metric, recoPrefix) {
	metric = recoMetricName
	metricDefExists = true
}
...
dbMetric := monitoredSource.Name + dbMetricJoinStr + metric ### interfere of dbMetric
_, cancelFuncExists := r.cancelFuncs[dbMetric] ### interfere of cancel functions
```

**Steps to reproduce**
1. run pgwatch and update the webui to use any `reco_*` metric with interval > 0
2. in the logs you will see that the metric name is changed to `recommendations` and interval is 0s.
```
2025-06-30 07:45:29.586 [WARNING] [source:books344] [metric:recommendations] Total fetching time of 92.563µs bigger than 0s interval
```

pashagolub:
- Remove hard-coded recommendations handling from reaper
- Add `metric_storage_name: "recommendations"` to all `reco_` metrics
- Create dedicated `recommendations` preset with distributed intervals
- Update `exhaustive/full` presets to use individual `reco_` metrics
- Add migration to handle existing installations
- Clean up legacy

### DIFF
--- a/internal/reaper/reaper.go
+++ b/internal/reaper/reaper.go
@@ -150,18 +150,17 @@ func (r *Reaper) Reap(ctx context.Context) {
 			hostLastKnownStatusInRecovery[monitoredSource.Name] = monitoredSource.IsInRecovery
 
 			for metricName, interval := range metricsConfig {
-				metric := metricName
 				metricDefExists := false
 				var mvp metrics.Metric
 
-				mvp, metricDefExists = metricDefs.GetMetricDef(metric)
+				mvp, metricDefExists = metricDefs.GetMetricDef(metricName)
 
-				dbMetric := monitoredSource.Name + dbMetricJoinStr + metric
+				dbMetric := monitoredSource.Name + dbMetricJoinStr + metricName
 				_, cancelFuncExists := r.cancelFuncs[dbMetric]
 
 				if metricDefExists && !cancelFuncExists { // initialize a new per db/per metric control channel
 					if interval > 0 {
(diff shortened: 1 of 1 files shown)


internal/reaper/recommendations.go
0xgouda, line 1: won't it be better to move the remaining special metrics consts definitions to reaper/metric.go and remove this file?
or at least rename it.

internal/reaper/reaper.go
```diff
@@ -146,12 +146,7 @@ func (r *Reaper) Reap(ctx context.Context) {
...
 
-				if strings.HasPrefix(metric, recoPrefix) {
-					metric = recoMetricName
-					metricDefExists = true
-				} else {
-					mvp, metricDefExists = metricDefs.GetMetricDef(metric)
-				}
+				mvp, metricDefExists = metricDefs.GetMetricDef(metric)
 
 				dbMetric := monitoredSource.Name + dbMetricJoinStr + metric
```
0xgouda: can't we now get rid of `metric` and use `metricName` in its place here and in other lines, to avoid confusion in the future?

PR #850, opened 2025-07-18
[!] migrate rpc sink to grpc

0xgouda:
- Add gRPC service definition `pgwatch.proto` file and its golang generated files to `api/pb/`
- Update `RPCWriter` to use gRPC client
- Use protobuf's `SyncOp` consts and remove old explicitly defined ones
- Remove `api/types.go` as all its types are now available at `api/pb/`


internal/sinks/multiwriter_test.go
```diff
@@ -37,9 +38,6 @@ func TestNewMultiWriter(t *testing.T) {
 		{&CmdOpts{
 			Sinks: []string{"prometheus://foo/"},
 		}, false, true},
-		{&CmdOpts{
-			Sinks: []string{"rpc://foo/"},
-		}, false, true},
```
0xgouda: If the gRPC client got passed `rpc://foo/` it won't give error until a call to a server method is invoked, and hence return nil error from `NewRPCWriter()` failing this test.

So if user entered wrong rpc server address he has to monitor the logs for errors, pgwatch won't fatally crash.

If this isn't acceptable i think we can work around it by invoking `SyncMetric("", "", InvalidOp)` call as a health check in `NewRPCWriter()` and ensure the error returned doesn't have `Unavailable` error code

pashagolub: Yeah, good idea to make a "ping" call! I like it

api/pb/pgwatch.proto
0xgouda, line 1: we shouldn't be removing `*pb*` files they have to exist, they have the definition for types/interfaces that are used in `rpc.go`.

it also should be the responsibility of the developer to compile them on change to `pgwatch.proto` not a workflow

api/pb/pgwatch.proto
```diff
@@ -25,8 +25,11 @@ message SyncReq {
...
+// SyncOp represents synchronization operations for metrics.
+// The actual constants are defined in internal/sinks/types.go
+// and should be used instead of these protobuf-generated constants.
 enum SyncOp {
-    InvalidOp = 0;
-    AddOp = 1;
-    DeleteOp = 2;
+    AddOp = 0;
+    DeleteOp = 1;
+    InvalidOp = -1;
```
0xgouda: Per the official docs negative numbers for protobuf enums aren't recommended.
```
Enumerator constants must be in the range of a 32-bit integer. 
Since enum values use varint encoding on the wire, 
negative values are inefficient and thus not recommended.
```
https://protobuf.dev/programming-guides/proto3/#:~:text=Enumerator%20constants%20must%20be%20in,inefficient%20and%20thus%20not%20recommended.

api/pb/pgwatch.proto
```diff
@@ -6,6 +6,7 @@ option go_package = ".;pb";
 service Receiver {
     rpc UpdateMeasurements(MeasurementEnvelope) returns (Reply);
     rpc SyncMetric(SyncReq) returns (Reply);
+    rpc DefineMetrics(google.protobuf.Struct) returns (Reply);
```
0xgouda, line 9: what about defining custom message type for `DefineMetric()` liks so.
```
pgwatch.proto
rpc DefineMetrics(Metrics) returns (Reply);
    
...
message Metrics {
    map<string, Metric> MetricDefs = 1;
    map<string, Preset> PresetDefs = 2;
}

message Metric {
    map<int32, string> SQLs = 1;
    string InitSQL = 2; 
    string NodeStatus = 3;
    repeated string Gauges = 4;
    bool IsInstanceLevel = 5;
    string StorageName = 6;
    string Description = 7;
}

message Preset {
    string Description = 1;
    map<string, double> Metrics = 2;
}
```

this can get us to avoid converting metrics to json and from json to`map[string]any` and then grpc have to do runtime checks to map `any` to `Struct` types, instead:
```
rpc.go/DefineMetrics()
grpcMetricDefs := make(map[string]*pb.Metric, len(metrics.MetricDefs))
for metricName, metricDef := range metrics.MetricDefs {
	grpcMetricDefs[metricName] = &pb.Metric{
	        SQLs: metricDef.SQLs,
	        InitSQL: metricDef.InitSQL,
		NodeStatus: metricDef.NodeStatus,
		Gauges: metricDef.Gauges,
		IsInstanceLevel: metricDef.IsInstanceLevel,
		StorageName: metricDef.StorageName,
		Description: metricDef.Description,
	}
}

grpcPresetDefs := make(map[string]*pb.Preset)
for key, val := range metrics.PresetDefs {
	grpcPresetDefs[key] = &pb.Preset{
		Description: val.Description,
		Metrics: val.Metrics,
	}
}

grpcMetrics := &pb.Metrics{
	MetricDefs: grpcMetricDefs,
	PresetDefs: grpcPresetDefs,
}	

reply, err := rw.client.DefineMetrics(rw.ctx, grpcMetrics)
```

this also will make better use of grpc message compression.

maybe we can do it in another PR this one already got too big.

internal/sinks/multiwriter_test.go
0xgouda: done, please check.

api/pb/pgwatch.proto
pashagolub: I don't see any benefits. This method executes only once at the start of the session. And most sinks will ignore it anyway. I started with a special custom message but then the amount of code made me mad and I used json as a straightforward solution. :)

api/pb/pgwatch.proto
pashagolub: Thanks! Didn't know that!

PR #861, opened 2025-07-24
[*] optimize maintenance tasks in Postgres sink, closes #786

### LINKED ISSUE #786: Optimize partition deletion workflow in `deleteOldPartitions()`
The current implementation of the `deleteOldPartitions()` function directly drops old partitions using `DROP TABLE`. This approach locks the parent table, potentially causing contention and delays in high-concurrency environments.

Objective is refactor the `deleteOldPartitions()` function to use `ALTER TABLE ... DETACH PARTITION CONCURRENTLY` before dropping partitions. This will reduce locking on the parent table and improve efficiency in partition management.to

pashagolub:
From now `admin.config` holds information about last successful operation. Maintenance routines will check if any other instance already processed it before starting it's own.

### DIFF
--- a/internal/sinks/postgres.go
+++ b/internal/sinks/postgres.go
@@ -127,6 +127,8 @@ type MeasurementMessagePostgres struct {
 	TagData map[string]string
 }
 
+const sourcesMaintenanceInterval = time.Hour * 24
+
 type DbStorageSchemaType int
 
 const (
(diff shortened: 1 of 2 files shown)


internal/sinks/postgres.go
```diff
@@ -534,40 +540,76 @@ func (pgw *PostgresWriter) deleteOldPartitions(delay time.Duration) {
...
 }
 
+func (pgw *PostgresWriter) StartMaintenanceActivity(ctx context.Context, activity string) bool {
+	logger := log.GetLogger(ctx)
+	tx, err := pgw.sinkDb.Begin(pgw.ctx)
+	if err != nil {
+		logger.Errorf("Starting transaction for %s maintenance failed: %w", activity, err)
+		return false
+	}
+	func() {
```
```suggestion
	defer func() {
```

pashagolub:
Overruled by #985

PR #862, opened 2025-07-24
[+] add basic auth support to grpc sink

0xgouda:
- send `username` and `password` contained in the grpc sink uri  `grpc://username:password@...` with each call's metadata to enable authentication for servers

### DIFF
--- a/internal/sinks/rpc.go
+++ b/internal/sinks/rpc.go
@@ -54,7 +54,7 @@ func NewRPCWriter(ctx context.Context, connStr string) (*RPCWriter, error) {
 		return nil, fmt.Errorf("error parsing gRPC URI: %s", err)
 	}
 
-	l := log.GetLogger(ctx).WithField("sink", "grpc").WithField("server", fmt.Sprintf("\"%s\"", uri.Host))
+	l := log.GetLogger(ctx).WithField("sink", "grpc").WithField("address", uri.Host)
 	ctx = log.WithLogger(ctx, l)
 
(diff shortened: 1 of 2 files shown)


internal/sinks/rpc.go
```diff
@@ -53,6 +54,9 @@ func NewRPCWriter(ctx context.Context, connStr string) (*RPCWriter, error) {
 		return nil, fmt.Errorf("error parsing gRPC URI: %s", err)
 	}
 
+	l := log.GetLogger(ctx).WithField("sink", "grpc").WithField("server", fmt.Sprintf("\"%s\"", uri.Host))
```
pashagolub: why do we want such a complex formatting?

0xgouda: It was about having the log be like
```
[server:"host:ip"] not [server:host:ip]
```

But if you think it is unnecessary or complicates things i can remove it

pashagolub: Yes, the reason is there are several output for logs, `--log-file-format=[json|text]`. So it's up to logger to format the output.
If we really want to change the formatting, we should apply changes to `internal/log/formatter.go`, in this case to `func (f *Formatter) writeField()`

pashagolub: I believe there is a reason not to use quotes. They eat up screen space :)

0xgouda: thanks for the insight, removed it.
but maybe i will play a bit with the logger in the future and see where i can get.

pashagolub (approved):
LGTM

PR #889, opened 2025-08-13
[+] allow metrics loading from folder, closes #882

0xgouda:
add `loadMetricsFromYaml()` to load metrics from file or folder of yaml files
  - read single yaml file or all yaml files in folder, unmarshal it/them, then copy it/them to the global `metrics.MetricDefs` and `metrics.PresetDefs` maps 
---
another PR that builds on top of this to make `Create[Metric|Preset]()`, `Update[Metric|Preset]()`, `Delete[Metric|Preset]()` functions handle metrics folder will be provided.        

closes: #882

PR #891, opened 2025-08-15
[+] implement [metric|preset] [create|delete|update] from metrics folder logic

### LINKED ISSUE #884: Can't add metric when adding PW_SOURCES as directory
Hello,

**Describe the bug**

I can't add any metric when I provide PW_SOURCES as a directory.

It may not be a bug, but it is not clear in the doc.

In my docker-compose.yaml file I have the following for my PW_SOURCES:

```yaml
[...]
  pgwatch:
    image: cybertecpostgresql/pgwatch:latest
    container_name: pgwatch
    environment:
      PW_SOURCES: "/config/pgwatch/sources/"
#      PW_METRICS: "/config/pgwatch/metrics/"
    command:
    - "--sink=postgresql://pgwatch:pgwatch@pgwatch_db:5432/pgwatch_metrics?sslmode=disable"
    - "--sink=prometheus://pgwatch:9187/metrics"
    ports:
      - "8080:8080"   # interface Web UI
      - "9187:9187"   # Prometheus exporter
    volumes:
      - "./pgwatch/sources/:/config/pgwatch/sources/"
      - "./pgwatch/metrics/:/config/pgwatch/metrics/"
    depends_on:
      pgwatch_db:
        condition: service_healthy
[...]
```

And when I add any metric, I get an "unsupported operation" message.

I think we should add a litle note about it, that when we use file based configuration we are not allowed to add any custom metrics from the UI. It makes sense in my opinion but not to everyone who whishes to put their hand on pgwatch.

Robin,

0xgouda:
- add `WalkMetricsDir(processMetricsFile func(metrics *Metrics) bool)` that traverses the metrics folder calling `processMetricsFile()` passing it a pointer to each file's metrics and if it returned true rewrite the file. 
  - this allows `update()` and `delete()` methods for presets and metrics to do their checks and updates on each file's metrics and decide whether it should be rewritten or not.
- update `Create[Metric|Preset]()` to write newly created [metrics|presets] to `new_metrics.yaml` file in case of using metrics folder
  - the file is created in `NewYAMLMetricReaderWriter()` on detecting folder configs.
  - this avoids the need to have existing yaml files in the folder to write to them
 ---
this builds on top of #889 

A new PR that builds on top of this to add tests and update docs will be provided.

For now this is not assumed to handle the scenario of (for some reason) having the same metric in different files in the metrics folder.

Note: if this is approved, we can take the same approach to resolve #884 its the exact the same update/delete issue for yaml folder but with sources instead.

### DIFF
--- a/internal/metrics/yaml.go
+++ b/internal/metrics/yaml.go
@@ -25,6 +25,8 @@ func NewYAMLMetricReaderWriter(ctx context.Context, path string) (ReaderWriter, 
 	isDir := false
 	if fi.Mode().IsDir() {
 		isDir = true
+		// create `new_metrics.yaml` to hold any 
+		// metrics/presets created via the web api
 		_, err = os.Create(filepath.Join(path, "new_metrics.yaml"))
 		if err != nil {
 			return nil, err
--- a/internal/metrics/yaml_test.go
+++ b/internal/metrics/yaml_test.go
@@ -68,6 +68,7 @@ func TestWriteMetricsToFile(t *testing.T) {
 	// Create a temporary file for testing
 	tempDir := t.TempDir()
 	tempFile := filepath.Join(tempDir, "metrics.yaml")
+	_, _ = os.Create(tempFile)
 
 	fmr, err := metrics.NewYAMLMetricReaderWriter(ctx, tempFile)
 	assert.NoError(t, err)
@@ -112,6 +113,7 @@ func TestMetricsToFile(t *testing.T) {
 	// Create a temporary file for testing
 	tempDir := t.TempDir()
 	tempFile := filepath.Join(tempDir, "metrics.yaml")
+	_, _ = os.Create(tempFile)
 
 	fmr, err := metrics.NewYAMLMetricReaderWriter(ctx, tempFile)
 	assert.NoError(t, err)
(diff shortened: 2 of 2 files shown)


internal/cmdopts/cmdconfig.go
```diff
@@ -59,6 +60,11 @@ func (cmd *ConfigInitCommand) InitSources() (err error) {
 func (cmd *ConfigInitCommand) InitMetrics() (err error) {
 	ctx := context.Background()
 	opts := cmd.owner
+
+	_, err = os.Create(opts.Metrics.Metrics)
```
0xgouda, line 64: This PR logic requires an already existing file or folder to be passed and then uses `os.Stat()` to differentiate between them, this will fail `config init` command if passed a non existing file.

The command supported only yaml files not folders  for bootstrapping configs, so this is a workaround that creates the file before running the command, maybe we can add some kind of support for bootstrapping yaml configs using either file or folder based approaches in the future.

pashagolub:
I don't want to go this way. pgwatch is already complex enough and I don't want to put a lot of efforts and make it full scale IDE. If people decided to split metrics into YAML files, then they used some other editor/IDE. And they have their own workflow for this.

So CRUD metric operations are available for database and single file configurations only. Maybe we should add some indicator to web UI signaling that user operates in RO mode.

Thanks anyway for your work!

PR #892, opened 2025-08-15
[*] update docs for metrics yaml folder configs

0xgouda:
- add note in docs about yaml folder based configs being read-only
- mention using yaml-folder for metrics configs in docs

### DIFF
--- a/docs/concept/installation_options.md
+++ b/docs/concept/installation_options.md
@@ -21,15 +21,16 @@ The default Docker demo image `cybertecpostgresql/pgwatch-demo` uses this approa
 ## File based operation
 
 One can deploy the gatherer daemon(s) decentralized with
-*sources to be monitored* defined in simple YAML files. In that case there
-is no need for the central Postgres configuration database. See the
+**sources to be monitored** and **metrics and presets definitions** 
+defined in simple YAML files. In that case there is 
+no need for the central Postgres configuration database. See
 [sample.sources.yaml](https://github.com/cybertec-postgresql/pgwatch/blob/master/contrib/sample.sources.yaml)
-config file for an example.
+And
+[sample.metrics.yaml](https://github.com/cybertec-postgresql/pgwatch/blob/master/contrib/sample.metrics.yaml)
+config files for an example.
 
 !!! Note
-    In this mode you also may want, but not forced, to point out the path to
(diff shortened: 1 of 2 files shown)


internal/cmdopts/cmdconfig.go
```diff
@@ -59,6 +60,11 @@ func (cmd *ConfigInitCommand) InitSources() (err error) {
 func (cmd *ConfigInitCommand) InitMetrics() (err error) {
 	ctx := context.Background()
 	opts := cmd.owner
+
+	_, err = os.Create(opts.Metrics.Metrics)
```
pashagolub: this shouldn't be here

internal/metrics/yaml.go
```diff
@@ -16,15 +16,40 @@ func NewYAMLMetricReaderWriter(ctx context.Context, path string) (ReaderWriter,
...
+		return nil, err
+	}
+
+	isDir := false
+	if fi.Mode().IsDir() {
+		isDir = true
+
+		// create `metrics_folder/new_metrics.yaml` if doesn't exist 
+		// to hold any metrics/presets created via the web api
+		newMetricsFilePath := filepath.Join(path, "new_metrics.yaml")
```
pashagolub: should be removed

pashagolub:
Since #891 is rejected this PR should be updated accordingly

0xgouda:
> Since #891 is rejected this PR should be updated accordingly

@pashagolub 

The PR has been updated, it now only updates docs for metrics yaml-folder configs, as tests for using yaml folder has already been provided #889 

Also noting that the [docs](https://pgwat.ch/latest/concept/installation_options.html#file-based-operation) mention env vars being usable for sources and metrics yaml setups, this logic is not implemented for metrics yaml configs, but idk if this is an actual bug or there is no need for supporting the use of envs in metrics yaml files and we only need to update the docs to mention that this is sources only feature?.

pashagolub:
Yeah, I cannot imagine how to use envvar in metric. So it's better to update docs

0xgouda:
> Yeah, I cannot imagine how to use envvar in metric. So it's better to update docs

done.

pashagolub (approved):

pashagolub:
Thanks! Well done!

Issue #764, opened 2025-05-20
Memory consumption - OOM


MichaelZhasil:
**Describe the bug**
With any amount of assigned memory there is always OOM and restart after a while. 
We have pgwatch running on several environments number of monitored databases vary from 10 to 100.
Currently we are running pgwatch on dedicated node with 54Gi of RAM assigned.
pgwatch start's working on about 1% of assigned capacity but within 10 minutes it hops up several times each hop consuming around 20% more of RAM until it fails on OOM and is restarted by K8S.
We switched to latest version 3.3.0 but the problem prevails 

**To Reproduce**
1. Have pgwatch running
2. Monitor tens of databases
3. Wait for a while - depends on assigned memory
4. OOM Killed
5. No visible error in pgwatch logs

**Expected behavior**
Memory consumption stops somewhere. / Some memory is freed

**Desktop (please complete the following information):**
 - Image version 3.3.0

**Additional context**
We are running full metrics preset, with only database sink.

0xgouda:
I tried to reproduce, can you confirm if this is the error you are getting?
```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xd0fa46]

goroutine 6849 [running]:
github.com/cybertec-postgresql/pgwatch/v3/internal/reaper.QueryMeasurements({0x1722758, 0xc000d13950}, {0xc000d8c208, 0x7}, {0xc00058af00, 0x2e8}, {0x0, 0x0, 0x0})
	/home/ahmed/pgwatch/internal/reaper/database.go:33 +0x206
github.com/cybertec-postgresql/pgwatch/v3/internal/reaper.(*Reaper).FetchMetric(0xc00038e3f0, {0x1722758, 0xc000d13950}, 0xc0008ec6c8, {0xc000c80720, 0x3}, 0xc000c0df48)
	/home/ahmed/pgwatch/internal/reaper/reaper.go:478 +0x92d
github.com/cybertec-postgresql/pgwatch/v3/internal/reaper.(*Reaper).reapMetricMeasurements(0xc00038e3f0, {0x1722758, 0xc000d13950}, 0xc0008ec6c8, {0xc000c80720, 0x3})
	/home/ahmed/pgwatch/internal/reaper/reaper.go:310 +0x919
created by github.com/cybertec-postgresql/pgwatch/v3/internal/reaper.(*Reaper).Reap in goroutine 1
[...]

PR #921, opened 2025-08-30
[+] improve recommendation metrics,  #764

pashagolub:
Refactor PostgreSQL recommendation metrics to consolidate multiple
findings per metric type.

The changes consolidate all 8 recommendation metrics to use
an aggregation pattern that combines multiple database objects into
comma-separated strings instead of separate rows, limiting results to
50 items with overflow indicators, and actionable guidance templates
were added to extra_info fields.

### DIFF
--- a/internal/metrics/metrics.yaml
+++ b/internal/metrics/metrics.yaml
@@ -19,7 +19,7 @@
 #        is_instance_level: true
 #        node_status: primary
 #        statement_timeout_seconds: 300
-#        metric_storage_name: db_stats
+#        storage_name: db_stats
 
 metrics:
     archiver:
@@ -485,7 +485,7 @@ metrics:
                   ) as catalog_size_b
         gauges:
             - '*'
-        metric_storage_name: db_size
+        storage_name: db_size
     db_stats:
         description: >
             Retrieves key statistics from the PostgreSQL `pg_stat_database` view, providing insights into the current database's performance.
@@ -748,7 +748,7 @@ metrics:
             - postmaster_uptime_s
             - backup_duration_s
             - checksum_last_failure_s
-        metric_storage_name: db_stats
+        storage_name: db_stats
     index_hashes:
         description: >
             Retrieves the hash of index definitions in the PostgreSQL database, providing a way to track changes in index definitions over time.
@@ -1553,7 +1553,7 @@ metrics:
         init_sql: "CREATE EXTENSION IF NOT EXISTS pg_qualstats;"
         node_status: primary
         is_private: true
-        metric_storage_name: "recommendations"
+        storage_name: "recommendations"
     reco_default_public_schema:
         description: >
             Retrieves recommendations for revoking the CREATE privilege on the public schema from PUBLIC.
@@ -1583,7 +1583,7 @@ metrics:
                 from rr, last_reco
                 where tag_object_name > '';
         node_status: primary
-        metric_storage_name: "recommendations"
+        storage_name: "recommendations"
     reco_disabled_triggers:
         description: >
             Retrieves recommendations for reviewing and potentially dropping disabled triggers in the PostgreSQL database.
@@ -1614,7 +1614,7 @@ metrics:
                 from rr, last_reco
                 where tag_object_name > '';
         node_status: primary
-        metric_storage_name: "recommendations"
+        storage_name: "recommendations"
     reco_drop_index:
         description: >
             Retrieves recommendations for dropping unused or invalid indexes in the PostgreSQL database.
@@ -1656,7 +1656,7 @@ metrics:
                 where tag_object_name > '';
         init_sql: "CREATE EXTENSION IF NOT EXISTS pg_qualstats;"
         node_status: primary
-        metric_storage_name: "recommendations"
+        storage_name: "recommendations"
     reco_nested_views:
         description: >
             Retrieves recommendations for overly nested views in the PostgreSQL database.
@@ -1732,7 +1732,7 @@ metrics:
                 from rr, last_reco
                 where tag_object_name > '';
         node_status: primary
-        metric_storage_name: "recommendations"
+        storage_name: "recommendations"
     reco_partial_index_candidates:
         description: >
             Retrieves recommendations for creating partial indexes on columns with a high fraction of NULL values.
@@ -1775,7 +1775,7 @@ metrics:
                     extra_info
                 from rr, last_reco
                 where tag_object_name > '';
-        metric_storage_name: "recommendations"
+        storage_name: "recommendations"
     reco_sprocs_wo_search_path:
         description: >
             Retrieves recommendations for stored procedures that do not have a fixed `search_path` set.
@@ -1807,7 +1807,7 @@ metrics:
                 from rr, last_reco
                 where tag_object_name > '';
         node_status: primary
-        metric_storage_name: "recommendations"
+        storage_name: "recommendations"
     reco_superusers:
         description: >
             Retrieves recommendations for reviewing the number of superusers in the PostgreSQL database.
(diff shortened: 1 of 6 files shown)


internal/metrics/metrics.yaml
```diff
@@ -1528,105 +1528,135 @@ metrics:
...
-                ORDER BY tag_object_name
+                with r as (select /* pgwatch_generated */
+                    (regexp_matches(v::text, E'ON (.*?) '))[1] as object_name,
+                    v::text as recommendation
+                    from json_array_elements(pg_qualstats_index_advisor() -> 'indexes') v),
+                rr as (select
+                    (extract(epoch from now()) * 1e9)::int8 as epoch_ns,
+                    'create_index'::text as tag_reco_topic,
+                    string_agg(object_name, ',') as tag_object_name,
+                    'create indexes for better query performance'::text as recommendation,
```
0xgouda, line 1539: won't this result in the string `create indexes for better query performance` be the result for all rows? 

the `recommendation` column in `r` CTE which has the real data is never accessed.

internal/metrics/metrics.yaml
```diff
@@ -1528,105 +1528,135 @@ metrics:
...
+                    'Use CREATE INDEX CONCURRENTLY <index_name> ON <table_name> (<columns>);'::text as extra_info
+                    from r 
+                    limit 50),
```
0xgouda, line 1542: Shouldn't this `LIMIT` be moved to `r` CTE instead, as `string_agg()` in `rr` will aggregate all results to a single row?

internal/metrics/metrics.yaml
```diff
@@ -1528,105 +1528,135 @@ metrics:
...
-                                'create_index'::text                       as tag_reco_topic,
-                                (regexp_matches(v::text, E'ON (.*?) '))[1] as tag_object_name,
-                                v::text                                    as recommendation,
-                                ''                                         as extra_info
-                         FROM json_array_elements(
-                                      pg_qualstats_index_advisor() -> 'indexes') v
-                     ) x
-                ORDER BY tag_object_name
+                with r as (select /* pgwatch_generated */
+                    (regexp_matches(v::text, E'ON (.*?) '))[1] as object_name,
```
0xgouda, line 1532: This will only match `<table-name>` from the recommendation text:

```
 select (regexp_matches('CREATE INDEX CONCURRENTLY idx_users_email ON users (email)', E'ON (.*?) '))[1];
 regexp_matches 
----------------
 users
 ```
 so the column name to create the index on won't be viewable to users.

File: internal/metrics/metrics.yaml (hunk 2 above)
Comment (0xgouda) [inline internal/metrics/metrics.yaml]: Shouldn't the `LIMIT` be moved to `r` CTE instead, as `string_agg()` in `rr` will aggregate all results to a single row anyway?

Example: It ignored the `limit 1` clause and aggregated all 3 emails.

```
select string_agg(email, ',') from (select email from Customer limit 3) limit 1;

                                             string_agg                                              
-----------------------------------------------------------------------------------------------------
 MARY.SMITH@sakilacustomer.org,PATRICIA.JOHNSON@sakilacustomer.org,LINDA.WILLIAMS@sakilacustomer.org
```

Suggestion: https://github.com/cybertec-postgresql/pgwatch/pull/921#discussion_r2312637510

Note:
This isn't the only metric that has this conflict between `string_agg()` and `LIMIT`, but I want to ensure that I am not getting something wrong first.

File: internal/metrics/metrics.yaml [hunk 4]
```diff
@@ -1528,105 +1528,135 @@ metrics:
...
+                    v::text as recommendation
+                    from json_array_elements(pg_qualstats_index_advisor() -> 'indexes') v),
+                rr as (select
+                    (extract(epoch from now()) * 1e9)::int8 as epoch_ns,
+                    'create_index'::text as tag_reco_topic,
+                    string_agg(object_name, ',') as tag_object_name,
+                    'create indexes for better query performance'::text as recommendation,
+                    'Use CREATE INDEX CONCURRENTLY <index_name> ON <table_name> (<columns>);'::text as extra_info
+                    from r
```
0xgouda: The `(regexp_matches(v::text, E'ON (.*?) '))[1]` will only match `<table-name>` from the recommendation text:

```
 select (regexp_matches('CREATE INDEX CONCURRENTLY idx_users_email ON users (email)', E'ON (.*?) '))[1];
 
regexp_matches 
----------------
 users
 ```
 So, with `r.recommendation` never being accessed in the query and only the `r.object_name` column is used in `rr` CTE.
The column name to create the index on won't be viewable to users, only the table name.

**Review by 0xgouda — COMMENTED** [review]
The regex only match `<table-name>` from the recommendation text:

```
 select (regexp_matches('CREATE INDEX CONCURRENTLY idx_users_email ON users (email)', E'ON (.*?) '))[1];
 
regexp_matches 
----------------
 users
 ```
 and `r.recommendation` is never accessed and only `r.object_name` column is used in `rr` CTE.
The column name to create the index on won't be viewable to users.

internal/metrics/metrics.yaml
0xgouda:
```suggestion
                limited_r as (
                  select * from r limit 50;
                ),
                rr as (select
                    (extract(epoch from now()) * 1e9)::int8 as epoch_ns,
                    'create_index'::text as tag_reco_topic,
                    string_agg(object_name, ',') as tag_object_name,
                    'create indexes for better query performance'::text as recommendation,
                    'Use CREATE INDEX CONCURRENTLY <index_name> ON <table_name> (<columns>);'::text as extra_info
                    from limited_r),
```

pashagolub: Oough! Nice catch! Thanks!

internal/metrics/metrics.yaml
pashagolub: well, yes, that's how it was written, I didn't change this. Now I'm thinking why do we truncate it at all...

0xgouda: > well, yes, that's how it was written, I didn't change this. Now I'm thinking why do we truncate it at all...

That's how it was written, but the first query had output:
```
tag_object_name          | recommendation
<table-name>             | "CREATE INDEX CONCURRENTLY idx_users_email ON users (email)"
```
Now we are only aggregating `object_name` and not using `r.recommendation` at all. 
idk if we should instead aggregate `r.recommendation` but that would be too long string, or update the regex to match table-name + column-name.

pashagolub: you're right! my bad. yeah, aggregate r.recommendation would be the easiest way. and for this metric we can limit the number of first recommendations to 10 maybe

internal/metrics/metrics.yaml
```diff
@@ -1528,105 +1528,147 @@ metrics:
...
+                limited_r as (select * from r limit 25),
+                rr as (select
+                    (extract(epoch from now()) * 1e9)::int8 as epoch_ns,
+                    'create_index'::text as tag_reco_topic,
+                    string_agg(object_name, ',') as tag_object_name,
+                    'create indexes for better query performance'::text as recommendation,
+                    'Use CREATE INDEX CONCURRENTLY <index_name> ON <table_name> (<columns>);'::text as extra_info
+                    from limited_r),
+                last_reco(s) as (
+                    select case when count(*) > 25 then format(' and found %s more...', count(*)-25) end from r),
```
0xgouda:
```suggestion
                limited_r as (select * from r limit 10),
                rr as (select
                    (extract(epoch from now()) * 1e9)::int8 as epoch_ns,
                    'create_index'::text as tag_reco_topic,
                    string_agg(object_name, ',') as tag_object_name,
                    string_agg(r.recommendation, ',') as recommendation,
                    'create these indexes for better query performance'::text as extra_info
                    from limited_r),
                last_reco(s) as (
                    select case when count(*) > 10 then format(' and found %s more...', count(*)-10) end from r),
```

internal/metrics/metrics.yaml
0xgouda: cool, suggestion: https://github.com/cybertec-postgresql/pgwatch/pull/921#discussion_r2313625486

PR #930, opened 2025-09-02
[!] remove obsolete `db_stats_aurora` metric

pashagolub:
The `db_stats_aurora` metric was created as a workaround for early AWS Aurora PostgreSQL limitations that prevented using certain PostgreSQL functions like pg_backup_start_time() and complex subqueries for invalid index detection.

Modern Aurora PostgreSQL versions (13+) now support virtually all standard PostgreSQL administration functions, making this Aurora-specific variant unnecessary. Users can simply use the standard `db_stats` metric instead, which provides the same functionality plus additional features like backup duration tracking and invalid index monitoring.

This removal reduces maintenance overhead and eliminates confusing

### DIFF
--- a/docs/howto/using_managed_services.md
+++ b/docs/howto/using_managed_services.md
@@ -27,58 +27,58 @@ such a use case.
 pgwatch has been tested to work with the following managed database
 services:
 
-# Google Cloud SQL for PostgreSQL
+## Google Cloud SQL for PostgreSQL
 
--   No Python / OS helpers possible. OS metrics can be integrated in
+- No Python / OS helpers possible. OS metrics can be integrated in
     Grafana though using the
     [Stackdriver](https://grafana.com/docs/grafana/latest/datasources/google-cloud-monitoring/)
     data source.
--   `pg_monitor` system role available.
--   pgwatch default preset name: `gce`.
--   Documentation: <https://cloud.google.com/sql/docs/postgres>
+- `pg_monitor` system role available.
+- pgwatch default preset name: `gce`.
+- Documentation: <https://cloud.google.com/sql/docs/postgres>
 
 To get most out pgwatch on GCE you need some additional clicks in the
 GUI / Cloud Console "Flags" section to enable some common PostgreSQL
 monitoring parameters like `track_io_timing` and `track_functions`.
 
-# Amazon RDS for PostgreSQL
+## Amazon RDS for PostgreSQL
 
--   No Python / OS helpers possible. OS metrics can be integrated in
+- No Python / OS helpers possible. OS metrics can be integrated in
     Grafana though using the
     [CloudWatch](https://grafana.com/docs/grafana/latest/datasources/cloudwatch/)
     data source
 
--   `pg_monitor` system role available.
+- `pg_monitor` system role available.
 
--   pgwatch default preset names: `rds`, `aurora`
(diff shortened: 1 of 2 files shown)


f9n:
Hi @pashagolub , 

Thanks for the cleanup! However, I'm running into an issue after the removal of the `aurora` preset. 

Since the preset was removed, I started using the `rds` preset for my AWS Aurora instances, but I'm now getting the following error regarding the `wal` metric:

`[ERROR] [metric:wal] [count:1] [error:ERROR: Function pg_last_xlog_replay_location() is currently not supported for Aurora (SQLSTATE 0A000)] failed to fetch metric data`

It seems that while Aurora 13+ supports most standard admin functions, standard WAL functions are still not supported due to its custom storage engine. 

I did some digging to see if we could bypass this in the SQL using a `CASE` statement. Would you consider updating the standard `wal` metric query to something like this to gracefully handle Aurora instances without needing a separate preset?

```sql
        wal:
            description: >
                This metric collects information about the Write-Ahead Logging (WAL) system in PostgreSQL.
                It provides insights into WAL activity, including the current WAL location, replay lag, and other related metrics.
            sqls:
                11: |-
                    select /* pgwatch_generated */
                    case
                        when exists (select 1 from pg_settings where name like 'aurora%') then null
                        when pg_is_in_recovery() = false then
                        pg_wal_lsn_diff(pg_current_wal_lsn(), '0/0')::int8
                        else
                        pg_wal_lsn_diff(pg_last_wal_replay_lsn(), '0/0')::int8
                        end as xlog_location_b,
                    case when pg_is_in_recovery() then 1 else 0 end as in_recovery_int,
                    extract(epoch from (now() - pg_postmaster_start_time()))::int8 as postmaster_uptime_s,
                    system_identifier::text as tag_sys_id,
                    case
                        when exists (select 1 from pg_settings where name like 'aurora%') then null
                        when pg_is_in_recovery() = false then
                        ('x'||substr(pg_walfile_name(pg_current_wal_lsn()), 1, 8))::bit(32)::int
                        else
                        (select min_recovery_end_timeline::int from pg_control_recovery())
                        end as timeline
                    from pg_control_system()
            gauges:
                - '*'
            is_instance_level: true
        wal_receiver:
            description: >
                This metric collects information about the WAL receiver process in PostgreSQL.
                It provides insights into the status of the WAL receiver, including replay lag and last replay timestamp.
            sqls:
                11: |-
                    select /* pgwatch_generated */
                      case
                        when exists (select 1 from pg_settings where name like 'aurora%') then null
                        else pg_wal_lsn_diff(pg_last_wal_receive_lsn(), pg_last_wal_replay_lsn())::int8
                      end as replay_lag_b,
                      case
                        when exists (select 1 from pg_settings where name like 'aurora%') then null
                        else extract(epoch from (now() - pg_last_xact_replay_timestamp()))::int8
                      end as last_replay_s
            node_status: standby
            gauges:
                - '*'
            is_instance_level: true
```

0xgouda:
@f9n 

I guess it's better to leave the metrics as they are, so it's clearer to new users that the `wal` and `wal_reciever` metrics are not compatible with Aurora, instead of just silently consuming storage while showing nothing useful on the dashboards.

But I will revert this and recreate an aurora preset that doesn't include `wal` and `wal_receiver`, does that work?

PR #957, opened 2025-09-19
feat: implement two-step partition cleanup for postgresql schema.

kylorend3r:
## PostgreSQL

- Replaced direct DROP TABLE with DETACH PARTITION CONCURRENTLY + DROP TABLE.
- Added getParentTableForPartition() helper method for parent table resolution.
- Updated `drop_old_time_partitions` to use detach + drop approach for PostgreSQL schema.

Screenshot 2025-09-20 at 15 51 54

## Timescale

- Introduced a two step chunk clean-up approach for timescale schema. Basically, it collects the hypertables and according to given interval it detach chunk. Afterwards, it drops the chunk/table.
- removed unused postgresql block inside the drop_old_time_partitions function because the function is being called to cleanup timescale schema only. 
Screenshot 2025-10-01 at 20 33 41

The purpose of this PR is to resolve issue described in #786

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/prometheus/client_golang v1.23.2
 	github.com/rifflock/lfshook v0.0.0-20180920164130-b9218ef580f5
 	github.com/sethvargo/go-retry v0.3.0
-	github.com/shirou/gopsutil/v4 v4.25.8
+	github.com/shirou/gopsutil/v4 v4.25.9
 	github.com/sirupsen/logrus v1.9.3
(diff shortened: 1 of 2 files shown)


pashagolub:
thanks for the PR! I will back to it after the official v4 release. Features for v4 are already frozen and I cannot add it to upcoming release.

kylorend3r:
@pashagolub your welcome, if it has still time to be merged I can work on timescale schema as well.  Let me work on this and update the PR.

pashagolub:
Overriden by #985

Issue #958, opened 2025-09-19
Add Configurable PostgreSQL Partition Intervals


kylorend3r:
## Purpose and Definition

Currently, PostgreSQL native partitioning in pgwatch uses a hardcoded weekly partition interval (1 week). This creates limitations for different deployment scenarios where partition size requirements vary based on:

```
      IF i = 0 THEN
          l_part_start := to_date(l_year::text || l_week::text, 'iyyyiw');
          l_part_end := l_part_start + '1week'::interval;
          part_available_from := l_part_start;
          part_available_to := l_part_end;
      ELSE
          l_part_start := l_part_start + '1week'::interval;
          l_part_end := l_part_start + '1week'::interval;
          part_available_to := l_part_end;
      END IF;

      l_part_name_3rd := format('%s_%s_y%sw%s', metric, dbname, l_year, to_char(l_week, 'fm00' ));

      IF char_length(l_part_name_3rd) > MAX_IDENT_LEN     -- use "dbname" hash instead of name for overly long ones
      THEN
          ideal_length = MAX_IDENT_LEN - char_length(format('%s__y%sw%s', metric, l_year, to_char(l_week, 'fm00')));
          l_part_name_3rd := format('%s_%s_y%sw%s', metric, substring(md5(dbname) from 1 for ideal_length), l_year, to_char(l_week, 'fm00' ));
      END IF;

      IF NOT EXISTS (SELECT 1
                      FROM pg_tables
                      WHERE tablename = l_part_name_3rd
                        AND schemaname = 'subpartitions')
      THEN
        l_sql := format($$CREATE TABLE IF NOT EXISTS subpartitions.%I PARTITION OF subpartitions.%I FOR VALUES FROM ('%s') TO ('%s')$$,
                        l_part_name_3rd, l_part_name_2nd, l_part_start, l_part_end);
        EXECUTE l_sql;
        EXECUTE format('COMMENT ON TABLE subpartitions.%I IS $$pgwatch-generated-metric-dbname-time-lvl$$', l_part_name_3rd);
```

- Data volume: High-volume monitoring setups need smaller partitions (daily)
- Storage constraints: Low-volume setups benefit from larger partitions (monthly)
- Query patterns: Different access patterns require different partition granularities
- Retention policies: Varying retention requirements affect optimal partition sizes

[...]

Issue #931, opened 2025-09-02
Partitions overlap when changing time zone of PostgreSQL


bkhomuts:
**Describe the bug**
After moving to a different PostgreSQL server, the following error occurred on 1st of September.
postgresql log:
```
ERROR:  partition "dbname_y2025w36" would overlap part
ition "backends_perun-mms.devk8s.efp.core.aai.geant.org_y2025w35" at character 198
pgwatch@pgwatch_metrics QUERY:  CREATE  TABLE IF NOT EXISTS subpartitions."backends_dbname_y2025w36" PARTITION OF subpartitions."dbname" FOR VALUES FROM ('2025-09-01') TO ('2025-09-08')
```
pgwatch log:
```
ERROR: no partition of relation "dbname" found for row (SQLSTATE 23514)
```

**To Reproduce**
1. At the end of the week, change timezone of the database from UTC to UTC+2.
2. On the first day of the next week, check the logs of pgwatch and postgresql. The data won't be recorded.

**Expected behavior**
Partition would be created cleanly.

**Additional context**
System: Debian 12, using pgwatch from the apt.postgresql.org repository
PGWatch 3.7.0-2.pgdg130+3

May be solved by checking for the end of last partition, and ensuring the new partition won't overlap overlap when creating it. It is possible also to create a partition with a specific timezone during partition creation, but this may result in error for existing installations.

PR #959, opened 2025-09-19
feat: Add CLI/Environment variable support for PostgreSQL partition intervals

kylorend3r:
### The Definition and Basic Details

- Introduced configurable partition intervals via the `--partition-interval` command-line option. It prevents year-based and minute-based intervals for optimal performance
- Implemented a partition naming strategy. For example, 
   - Daily partitions: metric_dbname_yyyymmdd
   - Weekly partitions: metric_dbname_y2024w01
   - Monthly partitions: metric_dbname_202401
   - Custom intervals: metric_dbname_20240101_1200_to_20240102_1200
- Added a new configuration to `admin.config` table to manage partition intervals for PostgreSQL type of schema in sink. 

### 1 Week (168h) Example
Screenshot 2025-10-17 at 09 26 00

### 3 Days (72h) Example

Screenshot 2025-10-17 at 09 05 18

### Possible Advantages

- Reduced lock contention during maintenance and more efficient partition dropping for data retention
- Ability to changing partition strategy via a config and a flexible approach to start with a custom interval depending on your requirements and environment.

Created this PR for #958


pashagolub:
Thanks for proposal but I don't like over engineering. 
We have `time.Duration` type in Go that perfectly describes intervals.

kylorend3r:
My goal was to introduce weekly,daily or monthly partitioned tables because currently we only create weekly partitions. For some cases single partitions can be too big or small and people may want to store them daily or monthly partitions. Thereby I wanted to introduce this option to be able to start with different partition intervals at the beginning. 

Let me know your opinions.

pashagolub:
What I am trying to say is if we want to provide custom intervals then why do you want to have predefined options? Let user decide what exactly interval they want. This way you don't need to bother with predefined options, just use time arithmetic and we're done

kylorend3r:
Hi @pashagolub , 

I had a chance to work on this MR according to your feedback and introduced the `partition-interval` for PostgreSQL schema type. 

Basically, the user will be able to define its own custom interval and partition naming will be handled in admin sql functions. I hope, thanks to this enhancements, people will be able to define custom partition intervals and change them for the new partitions to be created. 

Thanks for the feedback!

pashagolub:
Thanks for your work! I was working on the same issue. I will check what we have in 2 weeks. Until then Postgres conferences are in a full run

pashagolub:
Please check my approach to sql functions here f4cd404e94394901a4f0187479262c824fc31385. I'd prefer to use mine version because:
- there is no need in `admin.config` changes
- there are less checks all over the code for allowed/disallowed intervals
- names are simple and predictable
- `to_regclass` usage instead of complex queries
- the last partition bound used, which automatically solves #931 
- timescale partition handling improved and simplified a lot

kylorend3r:
Hi @pashagolub ,

- Does it make sense to have `admin.config` record to change the partition interval in the future ? I'd like to have it since it doesn't introduce performance penalty and provides some flexibility I suppose. This is just an idea to be more flexible.

I read the branch you mentioned and couldn't see a variable about `partition-interval` and `ensure_partition_metric_dbname_time` is not being called with interval by default,  is it ? 

- All validations are included in `cmdoptions.go` file I think it might be rigid to manage all validations from a single file and manage `admin.config`. The only validation is in this file. 

- We can convert naming to any proper format. 

In my humble opinion, we can proceed with the approach contains `admin.config` record. Let me know your thoughts, when it comes to #931 we can extend this branch to solve it. 

It was a pleasure to meet you in pgconf btw. Thanks for the support!

pashagolub:
hey! Nice to meet you in person! That was a cool event!

Probably, I didn't express my thoughts. :) 

Imagine you run your pgwatch (pseudo-code from here):

```sh
pgwatch --partition-interval=2d
```

Great, now you have your `admin.config = 2 days`. The very next day, we decided to run 
```sh
pgwatch --partition-interval=17h
```

What value will `admin.config` hold? What will happen to the old partitions? Whatever value hods `admin.config`, what is the point in it? That's wrong for partitions or it's wrong for new partitions.


Now, imagine I run 3 pgwatch instances with different `--partition-interval` values simultaneously. What value should hold `admin.config`? 

And the most important question: why do we need `admin.config` at all? pgwatch is responsible for maintaining partitions and it always knows the partition interval. Always. Why should we hold partition interval in `admin.config`? For what reasons?

In a branch I sent I wanted to show the approach of working with variable partition intervals. Function always finds the lower boundary and then creates the partition with interval that was passed to pgwatch via `--partition-interval`. This way we don't care about who, when and for how long created previous partition. We always create a new one based on existing.

I hope it's more or less clear.

kylorend3r:
From this point of view, it's clear and makes sense. 

Thanks for the examples,appreciated.

pashagolub:
Thanks for your contribution! I'm closing this PR as it was supersede by #1012

PR #980, opened 2025-10-03
[-] add `source` and `metric` tags to pgx traces, fixes #978

### LINKED ISSUE #978: Log level « error », get "source" info in each message
**Describe the bug**

Hi
I'm using « log-level=error » (otherwise there is too much messages), in combination with « --log-file-format=json » (which is the default), and I think it's missing informations on the source for some (most) of the errors.

Here is some examples of what I got : 

`{"args":[],"err":"ERROR: permission denied for function pg_stat_file (SQLSTATE 42501)","fields.time":24058773,"level":"error","msg":"Query","pid":1852202,"sql":"select /* pgwatch_generated */\n  (extract(epoch from now()) * 1e9)::int8 as epoch_ns, [...] ","time":"2025-10-02T14:43:53+02:00"}`

or 

`{"args":[],"err":"ERROR: function get_load_average() does not exist (SQLSTATE 42883)","fields.time":1739257,"level":"error","msg":"Query","pid":539080,"sql":"select /* pgwatch_generated */\n  (extract(epoch from now()) * 1e9)::int8 as epoch_ns,[...] ","time":"2025-10-02T14:43:49+02:00"}`

I tried using text format but it doesn't seem to be better (and queries are spanned over a bazillion lines)

**To Reproduce**
Classic usage of the product, my sources are in the source table in the database (500+ lines).

**Expected behavior**
To see "source":"my_source" on each log line, to be able to troubleshot errors targeting precisely the sources.

0xgouda:
- Resave logger to `ctx` after every `.WithField()` update, so that logs from called functions will use the same updated logger.  

closes: #978

### DIFF
--- a/internal/reaper/reaper.go
+++ b/internal/reaper/reaper.go
@@ -99,6 +99,7 @@ func (r *Reaper) Reap(ctx context.Context) {
 		hostsToShutDownDueToRoleChange := make(map[string]bool) // hosts went from master to standby and have "only if master" set
 		for _, monitoredSource := range r.monitoredSources {
 			srcL := logger.WithField("source", monitoredSource.Name)
+			ctx = log.WithLogger(ctx, srcL)
 
 			if monitoredSource.Connect(ctx, r.Sources) != nil {
 				srcL.Warning("could not init connection, retrying on next iteration")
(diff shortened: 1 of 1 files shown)


internal/reaper/database.go
```diff
@@ -464,8 +465,7 @@ func (r *Reaper) CheckForPGObjectChangesAndStore(ctx context.Context, md *source
 	md.Lock()
 	defer md.Unlock()
 	var err error
-	l := log.GetLogger(ctx).WithField("source", md.Name).WithField("metric", specialMetricChangeEvents)
-	ctx = log.WithLogger(ctx, l)
+	l := log.GetLogger(ctx).(*logrus.Entry)
```
pashagolub: do we really need explicit type casting here?

0xgouda: Yes we do.
In the below lines, `l.Data` and `l.String()` are only available under `logrus.Entry` which is what `.WithField()` used to return but now `GetLogger()` returns the raw `Logger` inteface so we have to cast. 
```
489 if len(l.Data) > 2 { // source and metric are always there
490		m := metrics.NewMeasurement(time.Now().UnixNano())
491		if m["details"], err = l.String(); err != nil {
		    ...
```

pashagolub (approved):
Perfect! What was needed!

PR #996, opened 2025-10-09
[*] make legends consistent in "Global Database Overview" dashboard

slardiere:
Harmonise the layout of legends: in a list and at the bottom for all signs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
       uses: ./.github/actions/setup-protobuf
 
     - name: Set up Node.js
-      uses: actions/setup-node@v5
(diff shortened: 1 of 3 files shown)


pashagolub:
would you please show before/after screenshots? It's hard to get the changes from the json definitions. thanks!

slardiere:
Of course, here's 2 screenshot : the actual, with 2 kind of legend, and my proposal, with the legend at the bottom, which give a fixed wide size, which i found easier to read : 
1-global-db-overview-actual
1-global-db-overview-proposal

( I can show you, because I'm in Riga too)

pashagolub:
Cool! Let's meet then! :)

PR #1011, opened 2025-11-06
[*] update `pgbench.sh` and `add-test-db.sh` file paths

0xgouda:
- `docker/compose.add-test-db.sh` has been moved to `docker/scripts/add-test-db.sh`.
- `docker/compose.pgbench.sh` has been moved to `docker/scripts/pgbench.sh`.

Update README accordingly.


pashagolub (approved):
Good catch!

PR #1012, opened 2025-11-06
[+] support variable partition intervals in postgres sink

0xgouda:
- Support specifying partition interval for PostgreSQL sink via new `--partition-interval` flag that accepts a [PostgreSQL interval](https://www.postgresql.org/docs/18/datatype-datetime.html#DATATYPE-INTERVAL-INPUT) string and uses it as the time range in the 3rd-level partition in metric tables.
- All formats for PostgreSQL intervals are supported.

closes: #958 
closes: #931


pashagolub:
Thanks. I don't see any new tests and I don't see validation on `--partition-interval` parameter.

I'd put additional check into `NewWriterFromPostgresConn` just as we did for `retention`.

pashagolub:
fixes #931 as well

Issue #994, opened 2025-10-08
Optimize maintainUniqueSources() load on shared sink databases


f9n:
**Is your feature request related to a problem? Please describe.**

Yes. When running `pgwatch` in a distributed setup — with multiple agents collecting metrics from different VMs and pushing data into a shared PostgreSQL sink — the `maintainUniqueSources()` function puts noticeable load on the sink database.

The recursive query that runs every 24h causes locking and contention, especially when there are tens of thousands of monitored sources.
We discussed a related topic in [#786](https://github.com/cybertec-postgresql/pgwatch/issues/786), but this one focuses specifically on optimizing `maintainUniqueSources()`.

**Describe the solution you'd like**

It would be great to have the option to:

Run maintenance tasks like `maintainUniqueSources()` and `deleteOldPartitions()` only on one designated node, e.g. via
`--admin-mode` or `--skip-maintenance-tasks`

Alternatively, make these tasks callable via subcommands, for example:
`pgwatch maintain` or `pgwatch cleanup`,
so they can be triggered manually or by a cronjob from a central controller node.

(Optional) Add a command like
`pgwatch cleanup --stale-sources --max-age=7d`
to remove sources that haven’t sent data recently, keeping the sink database clean.

**Describe alternatives you've considered**

We temporarily patched our internal build to reduce the frequency and impact of `maintainUniqueSources()`.
It works for now but isn’t ideal — long-term, having this built into pgwatch would make distributed deployments much cleaner and more efficient.

Internal references:
- https://github.com/f9n/pgwatch/releases/tag/v3.6.0%2Bty0.1.0
- https://github.com/f9n/pgwatch/commit/3786d66efd5a4e5bdd72fb00b399c54d0ff73028

**Additional context**

We currently manage over 52,000 unique dbname definitions, so running a single centralized pgwatch instance isn’t feasible due to metric lag.
Distributing the agents solved that issue but shifted maintenance overhead to the sink cluster.

Being able to delegate or externalize maintenance tasks would significantly improve scalability and reduce contention in large deployments.

cc: @emreyaniktr @selimtatlisu

pashagolub:
Thanks! We're on it!

PR #1018, opened 2025-11-11
[+] support custom configs for maintenance tasks

0xgouda:
- introduce `--maintenance-interval` flag that accepts a PostgreSQL interval string and controls the interval at which to run maintenance tasks. 0 disables it.
- modify `--retention` to accept PostgreSQL interval string instead of `time.Duration`, and use the zero value as an indicator to disable it as well. 
---
This **assumes** that migrating `--retention` from `time.Duration` to a `string` validated as a PostgreSQL interval will be backward-compatible.

closes: #994

### DIFF
--- a/internal/sinks/postgres.go
+++ b/internal/sinks/postgres.go
@@ -67,10 +67,9 @@ func NewWriterFromPostgresConn(ctx context.Context, conn db.PgxPoolIface, opts *
 			return err
 		}
 
-		// multiply by (10 ^ 9) because epoch returns seconds
-		// but time.Duration works in terms of nanoseconds
-		pgw.retentionInterval *= 1_000_000_000
-		pgw.maintenanceInterval *= 1_000_000_000
+		// epoch returns seconds but time.Duration represents nanoseconds
+		pgw.retentionInterval *= time.Second
+		pgw.maintenanceInterval *= time.Second
 
 		if !isValidPartitionInterval {
 			return fmt.Errorf("--partition-interval must be at least 1 hour, got: %s", opts.PartitionInterval)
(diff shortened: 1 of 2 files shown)


docs/reference/cli_env.md
```diff
@@ -98,17 +98,23 @@ It reads the configuration from the specified sources and metrics, then begins c
...
+    Time range for PostgreSQL sink table partitions. Must be a valid PostgreSQL interval. (default: 1 week)  
     ENV: `$PW_PARTITION_INTERVAL`
 
     Example:
     `--partition-inteval="3 weeks 4 days"`,
 
 - `--retention=`
 
-    If set, metrics older than that will be deleted (default: "14 days").  
+    Delete metrics older than this. set to zero to disable. Must be a valid PostgreSQL interval. (default: 14 days)
```
pashagolub:
```suggestion
    Delete metrics older than this. Set to zero to disable. Must be a valid PostgreSQL interval. (default: 14 days)  
```

internal/sinks/postgres.go
```diff
@@ -43,18 +58,28 @@ func NewWriterFromPostgresConn(ctx context.Context, conn db.PgxPoolIface, opts *
...
 		}
-		var isValidInterval bool 
-		err = conn.QueryRow(ctx, "SELECT $1::interval >= '1h'::interval", opts.PartitionInterval).Scan(&isValidInterval)
-		if err != nil {
-			return err
+
+		// multiply by (10 ^ 9) because we epoch returns seconds
+		// but time.Duration works in terms of nanoseconds
+		pgw.retentionInterval *= 1_000_000_000
+		pgw.maintenanceInterval *= 1_000_000_000
```
```suggestion
		pgw.retentionInterval *= time.Second
		pgw.maintenanceInterval *= time.Second
```

0xgouda: I guess `1_000_000_000` is easier for future readers to understand, as converting to nanoseconds via multiplying by `time.Second` may not be very intuitive at first glance?

pashagolub: using `time` constants is an idiomatic way. If I will see `pgw.retentionInterval *= 1_000_000_000` I will be confused a lot, frankly speaking.

> Another key benefit is type safety. Since time.Duration is its own type, the compiler will prevent you from accidentally mixing it with regular integers in ways that don't make sense. This helps avoid subtle bugs that plague time calculations in many applications.

https://dev.to/rezmoss/time-durations-and-calculations-in-go-310-maf

PR #1020, opened 2025-11-11
[*] fix maintenance task to delete entries of dropped tables

### LINKED ISSUE #1019: Dropped metric tables are not removed from `all_distinct_dbname_metrics` listing table
**Describe the bug**
If a metric table got dropped, its corresponding entries in `all_distinct_dbname_metrics` are not removed by the maintenance routine.

**To Reproduce**
1. Run pgwatch for some time with a disjoint set of metrics on each source and a small `--maintenance-interval`
2. Remove one source and drop its metric tables
3. Wait for the maintenance task to run 
3. query `select * from all_distinct_dbname_metrics` 
4. The dropped tables will still have entries in the listing table

**Expected behavior**
The entries for the dropped metric table should be removed from the listing table.

0xgouda:
builds on top of #1018 
closes #1019

### DIFF
--- a/internal/sinks/postgres.go
+++ b/internal/sinks/postgres.go
@@ -558,8 +558,8 @@ func (pgw *PostgresWriter) MaintainUniqueSources() {
 		rows, _ := pgw.sinkDb.Query(pgw.ctx, fmt.Sprintf(sqlDistinct, tableName, tableName))
 		ret, err := pgx.CollectRows(rows, pgx.RowTo[string])
 		if err != nil {
-			logger.Errorf("Could not refresh Postgres all_distinct_dbname_metrics listing table for '%s': %s", metricName, err)
-			break
+			logger.Errorf("Could not refresh Postgres all_distinct_dbname_metrics listing table for metric '%s': %s", metricName, err)
+			continue
 		}
 		for _, drDbname := range ret {
 			foundDbnamesMap[drDbname] = true // "set" behaviour, don't want duplicates


internal/sinks/postgres.go
```diff
@@ -520,6 +520,7 @@ func (pgw *PostgresWriter) MaintainUniqueSources() {
 	SELECT dbname FROM t WHERE dbname NOTNULL ORDER BY 1`
 	sqlDelete := `DELETE FROM admin.all_distinct_dbname_metrics WHERE NOT dbname = ANY($1) and metric = $2`
 	sqlDeleteAll := `DELETE FROM admin.all_distinct_dbname_metrics WHERE metric = $1`
+	sqlDroppedTables := `DELETE FROM admin.all_distinct_dbname_metrics WHERE NOT metric = ANY($1)`
```
```suggestion
	sqlDroppedTables := `DELETE FROM admin.all_distinct_dbname_metrics WHERE metric <> ANY($1)`
```

could this be just `<>`?

internal/sinks/postgres.go
```diff
@@ -545,13 +546,16 @@ func (pgw *PostgresWriter) MaintainUniqueSources() {
 		return
 	}
 
-	for _, tableName := range allDistinctMetricTables {
+	for i, tableName := range allDistinctMetricTables {
```
pashagolub, line 549: While the number of metric tables is less than 100 and additional loop shouldn't harm really, I don't like the idea of having additional loop just to trim `public.`.

**UPD: Ignore!**
~~If we really want that, I would rather update `sqlTopLevelMetrics` to something like sqlTopLevelMetrics := `SELECT (parse_ident(table_name))[2] FROM admin.get_top_level_metric_tables()`~~

internal/sinks/postgres.go
```diff
@@ -545,13 +546,16 @@ func (pgw *PostgresWriter) MaintainUniqueSources() {
...
+	}
+
+	for _, metricName := range allDistinctMetricTables {
 		foundDbnamesMap := make(map[string]bool)
 		foundDbnamesArr := make([]string, 0)
-		metricName := strings.Replace(tableName, "public.", "", 1)
 
 		logger.Debugf("Refreshing all_distinct_dbname_metrics listing for metric: %s", metricName)
-		rows, _ := pgw.sinkDb.Query(pgw.ctx, fmt.Sprintf(sqlDistinct, tableName, tableName))
+		rows, _ := pgw.sinkDb.Query(pgw.ctx, fmt.Sprintf(sqlDistinct, metricName, metricName))
```
pashagolub: This is a bug! You cannot use trimmed table name in `FROM` clause hoping that `public` will be the first schema in search path. Details: https://www.postgresql.org/docs/18/ddl-schemas.html#DDL-SCHEMAS-PATH

If attacker will execute something like
```sql
SET search_path TO myschema,"$user";
```
the code will query the tables from `myschema` only.

I don't understand what exactly you were trying to solve here?

internal/sinks/postgres.go
pashagolub: please, ignore. we need fully qualified table names. See the next comment.

pashagolub (changes requested):
Please, fix issues

internal/sinks/postgres.go
0xgouda: > I don't understand what exactly you were trying to solve here?

I was just trying to be extra cautious because if `sqlDroppedTables` got passed the tables without `public.*` prefix trimmed it will remove all the entries.

0xgouda: Also, as I said, this was **to be extra cautious** but actually for me, `allDistinctMetricTables` always gets returned with unqualified names lacking the `public.*` prefix, see below, so I think we need a modification to avoid the bug you mentioned?

Use `fmt.Println(allDistinctMetricTables)` result:
```
[configuration_changes index_changes object_changes privilege_changes sproc_changes table_changes wal]
```

pashagolub: aha, ok. So the thing is `admin.get_top_level_metric_tables()` knows exactly (sic!) how to return table names. In most case we will have the output as yours, because on most systems `public` is included in `search_path`

pashagolub: Let's do a call. We are talking about different things obviously

internal/sinks/postgres.go
```diff
@@ -520,6 +520,7 @@ func (pgw *PostgresWriter) MaintainUniqueSources() {
 	SELECT dbname FROM t WHERE dbname NOTNULL ORDER BY 1`
 	sqlDelete := `DELETE FROM admin.all_distinct_dbname_metrics WHERE NOT dbname = ANY($1) and metric = $2`
 	sqlDeleteAll := `DELETE FROM admin.all_distinct_dbname_metrics WHERE metric = $1`
+	sqlDroppedTables := `DELETE FROM admin.all_distinct_dbname_metrics WHERE metric != ALL($1)`
```
pashagolub, line 523: I'm not sure, can you check it?

```suggestion
	sqlDroppedTables := `DELETE FROM admin.all_distinct_dbname_metrics WHERE metric NOT IN ALL($1)`
```

That sounds like an idiomatic way

pashagolub (approved):
LGTM! thanks!

internal/sinks/postgres.go
0xgouda: It doesn't work; instead, it should be `NOT IN ($1)`, but it seems that pgx won't expand it into an array by default. I think the first one is good.

internal/sinks/postgres.go
```diff
@@ -545,17 +546,20 @@ func (pgw *PostgresWriter) MaintainUniqueSources() {
...
 
 		logger.Debugf("Refreshing all_distinct_dbname_metrics listing for metric: %s", metricName)
 		rows, _ := pgw.sinkDb.Query(pgw.ctx, fmt.Sprintf(sqlDistinct, tableName, tableName))
 		ret, err := pgx.CollectRows(rows, pgx.RowTo[string])
 		if err != nil {
-			logger.Errorf("Could not refresh Postgres all_distinct_dbname_metrics listing table for '%s': %s", metricName, err)
-			break
+			logger.Errorf("Could not refresh Postgres all_distinct_dbname_metrics listing table for metric '%s': %s", metricName, err)
+			continue
 		}
```
0xgouda, line 563: Used `continue` instead of `break` because:
1. Ensure we don't break from the loop before `before.*` is trimmed from all elements in `allDistinctMetricTables`.
2. There is no reason to break; we should continue and try to update entries for other metrics.
3. Not breaking at first failure and logging all metrics that would fail is more informative.

internal/sinks/postgres.go
pashagolub: From the manual:

```
expression operator ALL (array expression)
```

What error do you get with `NOT IN ALL($1)`?

pashagolub: ok, seems you are right! I messed subquery with array. 🤦

PR #1025, opened 2025-11-17
[*] migrate maintenance task logic to PL/pgSQL

0xgouda:
- Add new `admin.maintain_tables()` SQL function that abstracts the maintenance logic from the Go code
by handling the fetch of all metric tables, updating them one by one, and deleting entries for dropped tables.
---
TODO:
- [ ] Add sink schema migration


internal/sinks/postgres.go
```diff
@@ -509,95 +509,53 @@ func (pgw *PostgresWriter) DeleteOldPartitions() {
...
-		SELECT u, $2 FROM (select unnest($1::text[]) as u) x
-		WHERE NOT EXISTS (select * from admin.all_distinct_dbname_metrics where dbname = u and metric = $2)
-		RETURNING *`
-
 	var lock bool
-	logger.Infof("Trying to get admin.all_distinct_dbname_metrics maintainer advisory lock...") // to only have one "maintainer" in case of a "push" setup, as can get costly
-	if err := pgw.sinkDb.QueryRow(pgw.ctx, sqlGetAdvisoryLock).Scan(&lock); err != nil {
-		logger.Error("Getting admin.all_distinct_dbname_metrics maintainer advisory lock failed:", err)
+	logger.Infof("Trying to get maintenance advisory lock...") // to only have one "maintainer" in case of a "push" setup, as can get costly
+	if err := pgw.sinkDb.QueryRow(pgw.ctx, "SELECT admin.try_get_maintenance_lock();").Scan(&lock); err != nil {
```
pashagolub: I'm not sure it's a great improvement! :-D

0xgouda: It's not, but I think of it in terms of abstracting the maintenance logic from the Go code.

pashagolub: Got it. I will get back to this later. I want to experiment locally. Need to produce some loads first.

.github/workflows/build.yml
```diff
@@ -96,7 +96,7 @@ jobs:
     - name: Test
       run: |
         go generate ./api/pb/
-        go test -failfast -v -timeout=300s -p 1 -coverprofile=profile.cov ./cmd/... ./internal/...
+        go test -failfast -v -timeout=600s -p 1 -coverprofile=profile.cov ./cmd/... ./internal/...
```
pashagolub: 10 minutes? Why?

internal/sinks/sql/admin_functions.sql
```diff
@@ -139,3 +139,77 @@ BEGIN
   RETURN i;
 END;
 $SQL$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION admin.try_get_maintenance_lock(
```
pashagolub: Remove it. It gives nothing. Use `pg_try_advisory_lock()` directly.

pashagolub:
I want go part look like this:

```go
func (pgw *PostgresWriter) MaintainUniqueSources() {
	sql := "SELECT admin.maintain_tables() WHERE pg_try_advisory_lock(1571543679778230000)"
	logger := log.GetLogger(pgw.ctx)
	logger.Info("Starting maintainence...")
	if _, err := pgw.sinkDb.Exec(pgw.ctx, sql); err != nil {
		logger.Error("Maintaining measurement tables failed:", err)
	}
}
```

0xgouda:
> I want go part look like this:

done, please review.

.github/workflows/build.yml
0xgouda, line 99: Increased test timeouts to 10 minutes because `Test_MaintainUniqueSources_DeleteOldPartitions/MaintainUniqueSources` now takes more than 2 minutes alone, due to calls to `PG_SLEEP(60)` in `MaintainUniqueSources()`.

pashagolub: ping. anybody home? :-)

pashagolub:
Managed in #1100

PR #1029, opened 2025-11-18
[+] mention `RestartSec` and `TimeoutStopSec` for systemd

### LINKED ISSUE #1028: Improve Systemd File in Custom Installation Documentation
Hi, 

While reviewing the documentation I realised that we can use the following parameters while creating our systemd file

```
        RestartSec=5s
        TimeoutStopSec=60s
```

It will help other users to consider these parameters without having an issue while stopping the service. It's just a documentation tip it doesn't change the codebase&logic.

kylorend3r:
Added two parameters to the documentation.

Resolves #1028

### DIFF
--- a/docs/tutorial/custom_installation.md
+++ b/docs/tutorial/custom_installation.md
@@ -105,6 +105,8 @@ syntax differences.
         ExecStart=/usr/bin/pgwatch --sources=postgresql://pgwatch:xyz@localhost:5432/pgwatch --sink=postgresql://pgwatch:xyz@localhost:5432/pgwatch_metrics
         Restart=on-failure
         TimeoutStartSec=0
+        RestartSec=5s
+        TimeoutStopSec=60s
 
         [Install]
         WantedBy=multi-user.target


pashagolub:
nice! thanks for your input! should we put any commands on how to use those configs or do you think they are self-explanatory?

kylorend3r:
I believe they're widely known and self-explanatory ones. 

It would be pretty enough to add them to docs.

0xgouda:
@pashagolub 
I think it would be beneficial also to mention in custom_installations that pgwatch is now installable via apt from the [PostgreSQL APT repo](https://www.postgresql.org/download/linux/ubuntu/)?

The current apt version needs to be updated tho.

pashagolub:
> @pashagolub I think it would be beneficial also to mention in custom_installations that pgwatch is now installable via apt from the [PostgreSQL APT repo](https://www.postgresql.org/download/linux/ubuntu/)?
> 
> The current apt version needs to be updated tho.

Ooh, nice indeed! What about RPMs? That should go into a separate PR, I believe

PR #1053, opened 2025-11-28
[-] pin `pgwatch-metrics` data source to all panels and vars.

df7cb:
This:
1. Pins the `pgwatch-metrics` data source in all panels and vars in Grafana 12 postgres dashboards.
2. Removes dummy `"current": ...` tags to ensure users with problems in data sources won't see weird hard-coded values.
---
The same for the Grafana 11 postgres dashboard will be done in a separate PR to avoid having a huge number of changed files = more potential for unintentionally breaking smth.

### DIFF
--- a/grafana/postgres/v12/0-health-check.json
+++ b/grafana/postgres/v12/0-health-check.json
@@ -5041,10 +5041,6 @@
   "templating": {
     "list": [
       {
-        "current": {
-          "text": "demo",
-          "value": "demo"
-        },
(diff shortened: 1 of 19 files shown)


df7cb:
I did not bump any version numbers here, that should probably be done.

df7cb:
More dashboard with the same problem:
* $dbname in 4-tables-overview
* pgbouncer-stats: $monitoredsource

Issue #1046, opened 2025-11-27
Monitor JIT costs


ants:
JIT sometimes triggers for quick queries wasting CPU. It would be useful to see the fraction of time spent in JIT.

Should be collected from pg_stat_statements and then a health check / alert can check if a significant fraction of time is spent in JIT. jit_generation_time should be sufficient, other details are not particularly useful.

0xgouda:
Do you have any suggestions for thresholds to use, e.g., for green/orange/red alert flags?

PR #1054, opened 2025-11-28
[+] add `stat_statements_jit` metric

0xgouda:
- Update `stat_statements` metric to query `jit_generation_time`.
- Add a new `JIT %` panel to `0. Health Check` dashboard.
- Add a new `Top n by JIT %` to `Global Health` dashboard.
- Add a new `JIT %` column in `Query Performance Analysis` dashboard.

---
The `JIT %` represents the total amount of query execution time spent in JIT.

Closes: #1046

### DIFF
--- a/internal/metrics/metrics.yaml
+++ b/internal/metrics/metrics.yaml
@@ -3016,7 +3016,6 @@ metrics:
                     round(sum(temp_blk_write_time)::numeric, 3)::double precision as temp_blk_write_time,
                     sum(wal_fpi) as wal_fpi,
                     sum(wal_bytes) as wal_bytes,
-                    round(sum(s.jit_generation_time)::numeric, 3)::double precision as jit_generation_time,
                     round(sum(s.total_plan_time)::numeric, 3)::double precision as total_plan_time
                   from
                     pg_stat_statements s
@@ -3098,18 +3097,6 @@ metrics:
                     temp_blks_written desc
                   limit 100
                 ) a
-                union
-                select * from (
-                  select 
-                      *
-                  from
-                      q_data
-                  where
-                      jit_generation_time > 0
-                  order by
-                      jit_generation_time DESC
-                  limit 100
-                ) a; 
             17: |-
                 with /* pgwatch_generated */ q_data as (
                   select
@@ -3131,7 +3118,6 @@ metrics:
                     round(sum(temp_blk_write_time)::numeric, 3)::double precision as temp_blk_write_time,
                     sum(wal_fpi)::int8 as wal_fpi,
                     sum(wal_bytes)::int8 as wal_bytes,
-                    round(sum(s.jit_generation_time)::numeric, 3)::double precision as jit_generation_time,
                     round(sum(s.total_plan_time)::numeric, 3)::double precision as total_plan_time
                   from
                     pg_stat_statements s
@@ -3212,19 +3198,7 @@ metrics:
                   order by
                     temp_blks_written desc
                   limit 100
-                ) a
-                union
-                select * from (
-                  select 
-                      *
-                  from
-                      q_data
-                  where
-                      jit_generation_time > 0
-                  order by
-                      jit_generation_time DESC
-                  limit 100
-                ) a; 
+                ) a;
         storage_name: stat_statements
     subscription_stats:
         description: >


grafana/postgres/v12/0-health-check.json
```diff
@@ -5059,14 +5205,14 @@
         "type": "query"
       },
       {
-        "auto": false,
-        "auto_count": 30,
+        "auto": true,
+        "auto_count": 1,
```
0xgouda: 1. Added the  `auto` option to `online_interval` which: 
```
Dynamically calculates interval by dividing time range by the count specified
```
So count 1 gives us the same value as the current time range, thus decreasing the imprecision that could result from `$timeGroup(time, $__interval)` being divided into `5m, 15m, ...` groups.

---

Also, the final query values are not guaranteed to be 100% precise because we can't ensure that all the data will be placed in the same group across all interval-lengthened groups when using `$__timeGroup()`.

pashagolub: Is this a bug fix? If so, it should go to a separate PR. Here we work on jit only.

pashagolub:
We don't need another metric. We need updated `stat_statements` metric with a new column

0xgouda:
> We don't need another metric. We need updated `stat_statements` metric with a new column

That's what I did at first, but I noticed that `stat_statements_calls` metric exists and it directly aggregates the `calls` column instead of relying on the `calls` column in the `stat_statements` metric, so I did the same?

pashagolub:
The reason for a separate `stat_statements_calls` is simple:

```
            stat_statements: 180
            stat_statements_calls: 60
```

We want to run it every minute because it's lightweight. But everything else goes into `stat_statements`.

0xgouda:
> We want to run it every minute because it's lightweight. But everything else goes into `stat_statements`.

Yep, exactly; `stat_statements_jit` is almost identical to `stat_statements_calls`, just fetching a different field, so we may add it to presets (forgot to do so, on it) with a smaller interval than `stat_statements` as well. 

Also, if we just fetch another column in `stat_statements`, we will end up adding the union below to the query, which will complicate the query (increase the number of returned rows) and won't be very precise at the end.
```
UNION
    SELECT
        *
    FROM (
         SELECT
              *
          FROM
                q_data
          WHERE
                 jit_generation_time > 0
          ORDER BY
                jit_generation_time DESC
          LIMIT 100
     ) a
)
``` 
sounds okay?

pashagolub:
No. jit goes to stat_statements

pashagolub:
The thing is, if jit is not slow, we don't care

pashagolub:
Good question is: should we rewrite `stat_statements`? These `UNION` clauses look really ugly

pashagolub:
Another thing to think is plan related metrics. We need them as well

0xgouda:
> Another thing to think is plan related metrics. We need them as well

On my todo list.

0xgouda:
> No. jit goes to stat_statements

done, please review.

Also, if you think there are more useful panels for `JIT` than the ones added, please let me know.

internal/metrics/metrics.yaml
```diff
@@ -2986,6 +3016,7 @@ metrics:
                     round(sum(temp_blk_write_time)::numeric, 3)::double precision as temp_blk_write_time,
                     sum(wal_fpi) as wal_fpi,
                     sum(wal_bytes) as wal_bytes,
+                    round(sum(s.jit_generation_time)::numeric, 3)::double precision as jit_generation_time,
```
pashagolub: I don't think we want to duplicate it here

internal/metrics/metrics.yaml
```diff
@@ -2611,6 +2622,11 @@ metrics:
                       sum(wal_fpi)::int8 AS wal_fpi,
                       sum(wal_bytes)::int8 AS wal_bytes,
                       round(sum(s.total_plan_time)::numeric, 3)::double precision AS total_plan_time,
+                      round(sum(s.jit_generation_time)::numeric, 3)::double precision as jit_generation_time,
```
pashagolub, line 2625: Maybe we should get rid of this comment. We mentioned all metrics in docs, so I don't believe someone will get here to find this wisdom

0xgouda: afaik, in docs we didn't mention all metrics and mentioned `metrics.yaml` as source of truth, also metrics can be inspected via the webui, so I think it won't harm leaving it.

pashagolub:
why did you remove `jit` from `no_query` metric?

0xgouda:
> why did you remove `jit` from `no_query` metric?

Oh, Isn't that what you meant here :-) https://github.com/cybertec-postgresql/pgwatch/pull/1054#discussion_r2585749671?

PR #1056, opened 2025-12-01
[-] pin `pgwatch-metrics` data source to all v11 panels and vars

0xgouda:
This:

1. Pins the `pgwatch-metrics` data source in all panels and vars in Grafana 11 postgres dashboards.
2. Removes dummy `"current": ...` tags to ensure users with problems in data sources won't see weird hard-coded values.


pashagolub (approved):
Perfect!

PR #1057, opened 2025-12-01
[-] migrate all panels in `Health Check` dashboard to table format

0xgouda:
1. Remove `$online_interval` variable.
    - There is no reason for having a new time variable in addition to Grafana's time range.
3. Migrate all panels to `table format`.
    - panels had 3 approaches to add the time column required by `time series` format:
        1. `select 0 as time, ...`.
            - not necessary, remove `0 as time` and use `table format` instead.
        3. `select time, ....`.
            - same as above.
        5. `select $__timeGroup(time, $online_interval) .... group by 1`. 
            - not necessary as the data is already filtered by the user-specified time range using `$__timeFilter(time)`, so any further time grouping is not required, but actually can make the result imprecise, as some data that should be shown might fall into a different group that will be ignored.
3. Remove `where time > now() - '$online_interval'::interval`  and use `$__timeFilter(time)` instead.
---
TODO:
- [x] extra check on changed SQL queries to ensure I didn't break anything.


pashagolub:
Please, resolve conflicts. Then squash and merge. Thanks!

PR #1058, opened 2025-12-02
[-] fix sql of `Shared Buffers hit pct` panel

0xgouda:
The lines below in the `Shard Buffers hit pct` panel in the `0. Health Check` dashboard are unnecessary and decrease query performance:
```
(data->>'xact_rollback')::int8 as roll, lag((data->>'xact_rollback')::int8) over w as roll_lag,
(data->>'xact_commit')::int8 as comm, lag((data->>'xact_commit')::int8) over w as comm_lag,
```
---
It seems the query was copied as is from `TX rollback pct`, and those weren't removed.

### DIFF
--- a/grafana/postgres/v12/0-health-check.json
+++ b/grafana/postgres/v12/0-health-check.json
@@ -1020,7 +1020,7 @@
           "orderByTime": "ASC",
           "policy": "default",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


pashagolub (approved):
Good catch!

Issue #1059, opened 2025-12-02
Small values in `Overly long queries` panel.


0xgouda:
**Describe the bug**
The `Overly long queries` panel in the `Recommendation` dashboard shows a 1-second query as an overly long query.

<img width="718" height="97" alt="Image" src="https://github.com/user-attachments/assets/6e9f2337-f37e-4ce5-a34c-1fcd05610e8b" />
 

**To Reproduce**
1. Use the `backends` metric on your source
2. Run queries that will take a few seconds
3. wait for metric collection
4. Open the `Recommendations` dashboard under the pg source with the appropriate time range

**Expected behavior**
There should be a minimum for the long queries shown other than the `> 0` currently used.

0xgouda:
The `Longest query runtime` for any monitored database is already visible in `0. Health Check` Here we should be listing databases with truly `Overly long queries` only.

PR #1060, opened 2025-12-02
[-] fix recommendation dashboard panel name

0xgouda:
Closes: #1059


pashagolub:
queries **younger** then 30m? 👀 Did you mean shorter? Why 30m and not 1m?

PR #1063, opened 2025-12-02
[+] add support for reverse proxies on a different path, closes #1049

### LINKED ISSUE #1049: Support running behind a reverse proxy on a different path
I have pgwatch running on my server, but so far, I don't have access to the pgwatch webui because that's running on its own port (8432 here) that I don't want to expose directly to the internet.

For this, I have configured a reverse proxy in apache:
```
    ProxyPass /pgwatch/ http://localhost:8432/
    ProxyPassReverse /pgwatch/ http://localhost:8432/
    ProxyPreserveHost on
    <Location /pgwatch/>
        AuthUserFile /etc/apache2/admpasswd
        AuthType Basic
        AuthName "Administration"
        <RequireAll>
                Require valid-user
        </RequireAll>
    </Location>
```

The setup works, but the problem is that pgwatch is loading extra resources from / instead of /pgwatch:
```
<script defer="defer" src="/static/js/main.7c4fac15.js">
```

Grafana supports this configuration by letting the user specify a "root" directory:
```
# The full public facing url you use in browser, used for redirects and emails 
# If you use reverse proxy and sub path specify full url (with sub path) 
root_url = https://%(domain)s/grafana/
```

If pgwatch supported this as well, I would change the reverse proxy to talk to `http://localhost:8432/pgwatch/` and then things would work.

Please consider supporting this configuration, that's how web services are typically run in larger setups. Basically all web services of this kind support it.

pashagolub:
Users can now configure pgwatch to run under a custom path (e.g., http::/example.com/pgwatch) when behind a reverse proxy like Apache or Nginx.

- add `--web-base-path` flag to specify custom path prefix
- all API endpoints and static assets respect the configured base path
- inject `<base>` tag and React Router `basename` dynamically at runtime
- set `homepage: "."` in `package.json` for relative asset paths

Example usage:
  pgwatch --web-base-path=pgwatch --web-addr=:8080

Then configure reverse proxy to forward `/pgwatch/*` to `localhost:8080/pgwatch/*`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
       uses: actions/upload-artifact@v5
       with:
         name: webui-build
-        path: internal/webui/build
+        path: internal/webserver/build
 
     - name: GolangCI-Lint
       uses: golangci/golangci-lint-action@v9
@@ -88,7 +88,7 @@ jobs:
       uses: actions/download-artifact@v6
       with:
         name: webui-build
-        path: internal/webui/build        
+        path: internal/webserver/build        
 
     - name: Setup Protobuf
       uses: ./.github/actions/setup-protobuf


pashagolub:
@df7cb I'd really appreciate if you can test this one! Thanks in advance!

df7cb:
Thanks for the quick response. I tested the change and everything works like it should :heavy_check_mark: 

The only thing I would change is to fold the Apache WS configuration into the "main" Apache config snippet. The ProxyPass /pgwatch/log lines need to go before the /pgwatch ones or else the log proxy doesn't work. Perhaps there could be a "this is optional, leave out if you don't need the Logs tab" comment.

pashagolub:
@df7cb if you could show me how it should look like... Never worked with Apache

PR #1064, opened 2025-12-02
[+] add new `testutil` package

0xgouda:
- Add a new `testutil` package that contains helper functions for different tests. 
- Create a new `SetupPostgresContainer()` function that contains the pg container setup and termination logic.
- Move RPC tests servers setup, and mock Receiver to `testutil/rpc.go`.
- Create a separate `sinks/test_main.go` file containing the `TestMain()` for sinks tests.

### DIFF
--- a/internal/testutil/setup_test.go
+++ b/internal/testutil/setup_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/cybertec-postgresql/pgwatch/v3/api/pb"
 	"github.com/cybertec-postgresql/pgwatch/v3/internal/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -41,7 +40,7 @@ func TestLoadServerTLSCredentials(t *testing.T) {
 }
 
 func TestAuthInterceptor(t *testing.T) {
-	handler := func(ctx context.Context, req any) (any, error) {
+	handler := func(context.Context, any) (any, error) {
 		return "success", nil
 	}
 
@@ -130,9 +129,4 @@ func TestSetupPostgresContainer(t *testing.T) {
 	connStr, err := container.ConnectionString(context.Background())
 	require.NoError(t, err)
 	assert.NotEmpty(t, connStr)
-}
-
-func TestReceiverImplementsInterface(t *testing.T) {
-	// Compile-time check that Receiver implements pb.ReceiverServer
-	var _ pb.ReceiverServer = (*testutil.Receiver)(nil)
 }


pashagolub:
Great idea! Was on my todo list for a long time! :-) 

Let's go further! We need the same functions in many test packages anyway, not only sinks!

The idea is:
- add `/internal/testutil` special package
- put only `xxx_test.go` files in it (e.g. `docker_test.go`, `mock_test.go`), so we can sure it won't be ever used in production
- add SetupXXX and TearDownXXX functions, e..g
	```go
		func SetupPostgresDocker(t *testing.T) *Environment {
    		t.Helper()
			var err error
			const ImageName = "docker.io/postgres:17-alpine"
			pgContainer, err = postgres.Run(...)
	```
- each setup helper will return unique env for test, this way we are safe to run many tests in parallel

0xgouda:
> * each setup helper will return unique env for test, this way we are safe to run many tests in parallel

I am not sure I understand this point.

0xgouda:
> * put only `xxx_test.go` files in it (e.g. `docker_test.go`, `mock_test.go`), so we can sure it won't be ever used in production

This means we will be moving all our `*_test.go` files to this `/interna/testutil` dir, right?

pashagolub:
> I am not sure I understand this point.

If you setup a docker container, you want to return it to the test, right? If you setup a mock, you want the test to use it. That's what I mean under "return the environment". Every setup function should return test environment to the test case. No global vars or something like that.

pashagolub:
> This means we will be moving all our `*_test.go` files to this `/interna/testutil` dir, right?

No, `testutil` is exactly what the name says: "utilities to be used by tests".

For example instead of
```go
func TestReadMetricSchemaType(t *testing.T) {
	conn, err := pgxmock.NewPool()
	assert.NoError(t, err)
...
}
```

We could
```go
func TestReadMetricSchemaType(t *testing.T) {
	conn, tearDown := testutil.SetupMock()
	defer tearDown()
...
}
```


or in case of the docker:

```go
func TestFoo(t *testing.T) {
	pg, tearDown := testutil.SetupPgContainer()
	defer tearDown()
	connStr, err = pg.ConnectionString(ctx)
	assert.NoError(t, err)
	assert.NoError(t, db.Ping(ctx, connStr))
}
```


Setup returns everything test case needs to run (environment) and the `tearDown()` function that test calls at the end to free resources, cleanup temp files, etc.

internal/testutil/setup.go
```diff
@@ -0,0 +1,108 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+var ctx = log.WithLogger(context.Background(), log.NewNoopLogger())
+
+func SetupPostgresContainer() (*postgres.PostgresContainer, func(), error) {
+	const ImageName = "docker.io/postgres:17-alpine"
```
pashagolub: should it go to `helper_vars.go`?

internal/testutil/setup.go
```diff
@@ -0,0 +1,108 @@
...
+	pgContainer, err := postgres.Run(ctx,
+		ImageName,
+		postgres.WithDatabase("mydatabase"),
```
pashagolub: should "mydatabase" be the package var, so in tests we can
```go
logger.Info("Connected to", testutil.MockDbName)
```

internal/testutil/setup.go
```diff
@@ -0,0 +1,108 @@
...
+
+	clientUsername := md.Get("username")[0]
+	clientPassword := md.Get("password")[0]
+
+	if clientUsername != "" && clientUsername != "pgwatch" && clientPassword != "pgwatch" {
+		return nil, status.Error(codes.Unauthenticated, "unauthenticated")
+	}
+
+	return handler(ctx, req)
+}
```
pashagolub, line 63: I believe if something is used only once, and gRPC methods and structs are used only in one package, nowhere else, it should be local. So I'd return these back.

`testutil` is only for functionality that shared across different packages.

internal/sinks/postgres_test.go
```diff
@@ -7,19 +7,16 @@ import (
...
 	"github.com/pashagolub/pgxmock/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	testcontainers "github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/modules/postgres"
-	"github.com/testcontainers/testcontainers-go/wait"
 )
 
-var ctx = log.WithLogger(context.Background(), log.NewNoopLogger())
+var pgContainer *postgres.PostgresContainer
```
pashagolub: It shouldn't be global. Remember I told you about parallel execution of tests?

internal/sinks/postgres_test.go
```diff
@@ -461,19 +458,9 @@ func TestCopyFromMeasurements_StateManagement(t *testing.T) {
...
-		testcontainers.WithWaitStrategy(
-			wait.ForLog("database system is ready to accept connections").
-				WithOccurrence(2).
-				WithStartupTimeout(5*time.Second)),
-	)
-	r.NoError(err)
-	defer func() { assert.NoError(t, pgContainer.Terminate(ctx)) }()
 
-	connStr, _ := pgContainer.ConnectionString(ctx, "sslmode=disable")
+	connStr, err := pgContainer.ConnectionString(ctx, "sslmode=disable")
```
pashagolub, line 467: Where is initialization of the container?

internal/sinks/postgres_test.go
```diff
@@ -513,21 +500,9 @@ func TestCopyFromMeasurements_CopyFail(t *testing.T) {
...
-		testcontainers.WithWaitStrategy(
-			wait.ForLog("database system is ready to accept connections").
-				WithOccurrence(2).
-				WithStartupTimeout(5*time.Second)),
-	)
-	r.NoError(err)
-	defer func() { a.NoError(pgContainer.Terminate(ctx)) }()
 
-	connStr, _ := pgContainer.ConnectionString(ctx, "sslmode=disable")
+	connStr, err := pgContainer.ConnectionString(ctx, "sslmode=disable")
```
pashagolub, line 513: Where is init?

internal/sinks/main_test.go
pashagolub, line 1: I don't want TestMain

internal/testutil/setup.go
0xgouda: yeah maybe.
moved it in #1068 to avoid rebasing much stuff.

pashagolub:
One more thing. I don't like `helper_vars.go` name. It might be just `types.go`, the same way as in other packages

internal/sinks/main_test.go
0xgouda: I thought it would make tests faster, instead of creating the pgcontainer 5 times or so.

internal/testutil/setup.go
0xgouda: > I believe if something is used only once, and gRPC methods and structs are used only in one package

Actually, in all tests, the only logic shared between different packages is `SetupPostgresContainer()`, but I don't think that `testutil` should contain only that?

0xgouda: I mostly moved all helper functions and mocks to `testutil`, as I think this would make their original tests more readable and maintainable.

internal/sinks/postgres_test.go
0xgouda: It seems that `testcontainers` tends to reuse the same container even when initialized in different variables (I tested this before), so it's the same, and hence our tests should already work in parallel, even with a global variable, but let me extra check on that.

0xgouda: I don't think this will matter for parallelism, since `testcontainers` already reuses containers sometimes. But I think your approach is cleaner anyway.

internal/sinks/rpc_test.go
```diff
@@ -1,146 +1,42 @@
-package sinks_test
+package sinks
```
pashagolub: any reason not to use `sinks_test`?

0xgouda: Oh, when we had `TestMain()` in `main_test.go`, everything should have been in the same package, now I can revert it.

0xgouda: But anyway, being in `sink` or `sinks_test` won't matter; it's the same either way.

pashagolub: it's more a matter of style. if you use a dedicated package it shows that you have clear separation. :)

pashagolub (approved):
+1

PR #1067, opened 2025-12-04
[*] set lock_timeout via connection RuntimeParams instead of per-query transaction

NikolayS:
Reduces metric queries from 4 roundtrips (BEGIN/SET/query/COMMIT) to 1 by setting lock_timeout at connection level via pgx RuntimeParams.

- Add --conn-lock-timeout option (default: 100ms, env: PW_CONN_LOCK_TIMEOUT)
- Set lock_timeout in RuntimeParams for PostgreSQL sources on connect
- Simplify QueryMeasurements by removing transaction wrapper
- Set to "0" to disable lock_timeout

Ref: https://brandur.org/fragments/postgres-parameters


pashagolub:
Thanks! Can you ask Claude to fix tests or should I? :-)

NikolayS:
> Thanks! Can you ask Claude to fix tests or should I? :-)

@pashagolub attempted in https://github.com/cybertec-postgresql/pgwatch/pull/1067/commits/56d368f0892904f2e39668d276ebf57da006957d

PR #1068, opened 2025-12-04
[*] move all test helpers to `testutil`

0xgouda:
- This builds on top of #1064 and moves all remaining test helpers to the new `testutil` package.


0xgouda:
overtaken by #1071

PR #1071, opened 2025-12-05
[*] migrate all helpers to `testutil`

0xgouda:
- builds on top of #1064 and migrates all remaining helpers/mocks to `testutil` package.
---
TODO:
- [x] Fix race condition that may originate from using `SetupRPCServers()` in tests multiple times. 
- [x] Add more tests to increase test coverage.

### DIFF
--- a/grafana/postgres/v12/0-health-check.json
+++ b/grafana/postgres/v12/0-health-check.json
@@ -19,7 +19,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 1,
+  "id": 21,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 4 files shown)


internal/db/bootstrap_test.go
```diff
@@ -25,8 +22,10 @@ func TestPing(t *testing.T) {
 	connStr := "foo_boo"
 	assert.Error(t, db.Ping(ctx, connStr))
 
-	pg, err := initTestContainer()
+	pg, pgTeardown, err := testutil.SetupPostgresContainer()
```
pashagolub, line 25: How do you think maybe we shouldn't return `err` from `testutil.SetupPostgresContainer`? Instead just `panic()` in case of error

pashagolub: or we can define it as `testutil.SetupPostgresContainer(t *testing.T)` and call `t.Fatal()` in case of the error

0xgouda: Would it differ? Also, I believe it's a global pattern to return the error to the first caller, and then he can decide whether to panic or ignore the error.

internal/reaper/database_test.go
```diff
@@ -13,10 +13,8 @@ import (
 )
 
 // Helper function to create a test SourceConn with pgxmock
-func createTestSourceConn(t *testing.T) (*sources.SourceConn, pgxmock.PgxPoolIface) {
+func createTestSourceConn() (*sources.SourceConn, pgxmock.PgxPoolIface, error) {
```
pashagolub: Revert this

internal/reaper/reaper_test.go
```diff
@@ -10,27 +10,11 @@ import (
...
-		return nil, m.toErr
-	}
-	return m.toReturn, nil
-}
-
-func (m *mockReader) WriteSources(sources.Sources) error { return nil }
-func (m *mockReader) DeleteSource(string) error          { return nil }
-func (m *mockReader) UpdateSource(sources.Source) error  { return nil }
-func (m *mockReader) CreateSource(sources.Source) error  { return nil }
-
```
pashagolub, line 33: Revert this. `testutil` **must** contain only structs and functions that are used in more than one place.

internal/testutil/setup.go
```diff
@@ -35,6 +36,37 @@ func SetupPostgresContainer() (*postgres.PostgresContainer, func(), error) {
...
+
+	tearDown := func() {
+		_ = pgContainer.Terminate(ctx)
+	}
+
+	return pgContainer, tearDown, err
+}
+
+func SetupEtcdContainer() (*etcd.EtcdContainer, func(), error) {
+	etcdContainer, err := etcd.Run(ctx, "gcr.io/etcd-development/etcd:v3.5.14",
```
pashagolub: add a const for the etcd version just like for postgres

internal/testutil/mocks.go
```diff
@@ -41,3 +44,73 @@ func (receiver *Receiver) DefineMetrics(_ context.Context, metricsStruct *struct
...
+	return m.WriteSourcesFunc(srcs)
+}
+
+type MockFS struct {
+	OpenFunc func(name string) (fs.File, error)
+}
+
+func (m MockFS) Open(name string) (fs.File, error) {
+	return m.OpenFunc(name)
+}
```
pashagolub: Revert this. `testutil` **must contain only structs and functions that are used in more than one place.**

internal/webserver/source_test.go
```diff
@@ -10,35 +10,12 @@ import (
...
-	return m.CreateSourceFunc(md)
-}
-func (m *mockSourcesReaderWriter) DeleteSource(name string) error {
-	return m.DeleteSourceFunc(name)
-}
-func (m *mockSourcesReaderWriter) WriteSources(srcs sources.Sources) error {
-	return m.WriteSourcesFunc(srcs)
-}
-
-func newTestSourceServer(mrw *mockSourcesReaderWriter) *WebUIServer {
```
pashagolub, line 41: Revert this. `testutil` **must** contain only structs and functions that are used in more than one place.

internal/webserver/webserver_test.go
```diff
@@ -11,17 +11,11 @@ import (
...
 	"github.com/stretchr/testify/assert"
 )
 
-type mockFS struct {
-	OpenFunc func(name string) (fs.File, error)
-}
-
-func (m mockFS) Open(name string) (fs.File, error) {
-	return m.OpenFunc(name)
-}
```
pashagolub, line 24: Revert this. `testutil` **must** contain only structs and functions that are used in more than one place.

internal/webserver/metric_test.go
```diff
@@ -10,47 +10,12 @@ import (
...
-	return m.UpdatePresetFunc(name, preset)
-}
-func (m *mockMetricsReaderWriter) CreatePreset(name string, preset metrics.Preset) error {
-	return m.CreatePresetFunc(name, preset)
-}
-func (m *mockMetricsReaderWriter) WriteMetrics(metricDefs *metrics.Metrics) error {
-	return m.WriteMetricsFunc(metricDefs)
-}
-
-func newTestMetricServer(mrw *mockMetricsReaderWriter) *WebUIServer {
```
pashagolub, line 53: Revert this. `testutil` **must** contain only structs and functions that are used in more than one place.

pashagolub:
It's probably easier to start a new PR. I propose to focus **only** on docker container helpers first.

internal/webserver/source_test.go
0xgouda: `mockSourcesReaderWriter` is already used in more than one place. If you observe the changes at `reaper_test.go`, you will see that I reused it there and removed the other sources mock that existed

internal/reaper/reaper_test.go
0xgouda: This sources mock was removed and I used the general sources mock defined previously in `webserver` (moved to `testutil`) instead. 
refer to https://github.com/cybertec-postgresql/pgwatch/pull/1071#discussion_r2599362671

internal/testutil/mocks.go
0xgouda: 1. Reverted `MockFS`.
2.  `MockSourcesReaderWriter` is already used in multiple places as mentioned in  https://github.com/cybertec-postgresql/pgwatch/pull/1071#discussion_r2599369751, 
3. For `mockMetricsReaderWriter`, I think it should stay, as it's very probable that it will get used in the future in packages other than `webserver`.
---

Now in this PR we have only migrated container helpers, and these `Mock{Metrics,Sources}ReaderWriter` mocks.

Issue #1070, opened 2025-12-05
Remote Log Parsing via PostgreSQL Server Functions


pashagolub:
## Problem Statement

Currently, pgwatch can only parse PostgreSQL server logs when it runs on the same host as the monitored database. This requires direct filesystem access to log files. This limitation prevents monitoring remote PostgreSQL instances for log-based metrics like `server_log_event_counts`.

## Solution Overview

Enable pgwatch to parse PostgreSQL server logs remotely using PostgreSQL's Generic File Access Functions. These functions are available to users with the `pg_monitor` role (PostgreSQL 10+) and allow reading log files through SQL queries instead of direct filesystem access.

## Key PostgreSQL Functions to Use

- `pg_ls_logdir()` - List log files in the log directory
- `pg_read_file(filename, offset, length)` - Read portions of text files
- `pg_stat_file(filename)` - Get file metadata (size, modification time)

### 1. Mode Detection and Selection

Create automatic detection logic that:
- Checks if pgwatch is on the same host as PostgreSQL (prefer local for performance)
- Tests if remote functions are available and accessible (PostgreSQL 10+, proper permissions)
- Falls back gracefully if remote access is not possible

### 2. Remote File Discovery

Replace filesystem operations with SQL-based equivalents:
- Use `pg_ls_logdir()` to list available log files (replaces `filepath.Glob()`)
- Use `pg_stat_file()` to get file sizes and modification times (replaces `os.Stat()`)
- Identify the latest log file based on modification timestamp

### 3. Remote File Reading

Implement reading log content through SQL queries:
- Track read position (byte offset) in memory between intervals
- Use `pg_read_file(path, offset, length)` to read new content since last check
- Parse CSV log lines using existing regex logic
- Handle log rotation by detecting new files and resetting position

### 4. Refactor Main ParseLogs Function

Split current implementation into two paths:
- `parseLogsLocal()` - Current filesystem-based approach (renamed, unchanged)
- `parseLogsRemote()` - New SQL-based approach
- Main `ParseLogs()` function detects mode and calls appropriate implementation

### 5. Configuration

Maintain backward compatibility (no config changes required for existing setups)

### 6. Documentation Updates

Update user documentation to cover:
- Remote log parsing capability and requirements
- Required permissions (`pg_monitor` role)
- Performance considerations and recommended intervals
- Limitations and supported log formats

### Unit Tests
[...]

PR #1072, opened 2025-12-09
[!] support parsing remote logs

0xgouda:
- Support parsing event counts from remote sources' logs.

TODO:
- [x] Add Unit & Integration tests.
- [x] Update docs.
- [x] Test functionality for a while on staging VM.
- [x] Benchmark estimated latency. 
    - about 7s additional overhead for every 10MBs (will differ based on single record size, record used = `raise warning 'I am warning'`)
- [x] Double check backward compatibility.
- [x] Move to a new `pglogparse` package (optional).
- [x] Set max read size (10 MB).

Closes: #1070

### DIFF
--- a/internal/reaper/logparser_test.go
+++ b/internal/reaper/logparser_test.go
@@ -10,12 +10,11 @@ import (
 
 	"github.com/cybertec-postgresql/pgwatch/v3/internal/metrics"
 	"github.com/cybertec-postgresql/pgwatch/v3/internal/sources"
+	"github.com/cybertec-postgresql/pgwatch/v3/internal/testutil"
 	"github.com/pashagolub/pgxmock/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
-
-var testCtx = context.Background()
 
 func TestNewLogParser(t *testing.T) {
 	tempDir := t.TempDir()
@@ -26,7 +25,8 @@ func TestNewLogParser(t *testing.T) {
 
 	sourceConn := &sources.SourceConn{
 		Source: sources.Source{
-			Name: "test-source",
+			Name:    "test-source",
+			Metrics: map[string]float64{specialMetricServerLogEventCounts: 60.0},
 		},
 		Conn: mock,
 	}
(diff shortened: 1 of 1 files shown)


0xgouda:
@pashagolub PR is Ready for review.

---

Regarding the below point in https://github.com/cybertec-postgresql/pgwatch/issues/1070:
> 8. Cloud-Ready: Works with managed PostgreSQL services (AWS RDS, Azure Database)

I haven't tested that.

internal/logparse/logparse.go
```diff
@@ -0,0 +1,182 @@
...
+		return
+	}
+	logger.Debugf("Using %s as log parsing regex", logsRegex)
+
+	if LogsGlobPath != "" {
+		logsGlobPath = LogsGlobPath
+	} else {
+		if logsGlobPath, err = tryDetermineLogFolder(ctx, mdb.Conn); err != nil {
+			logger.WithError(err).Print("Could not determine Postgres logs folder")
+			return
```
pashagolub: shouldn't we fall back to remote parsing instead of return?

pashagolub:
about naming. if we really want a separate package it should be `logparser`

pashagolub:
Instead of having three `ParseLogs` implementation, I'd really love to see
```go 

type LogParser struct{...}

func NewLogParser(mdb *sources.SourceConn, storeCh chan<- metrics.MeasurementEnvelope) *LogParser {...}

func (lp *LogParser) ParseLogs(context.Context) {...}

func (lp *LogParser) ParseLocal(context.Context) {...}

func (lp *LogParser) ParseRemote(context.Context) {...}
```

this way we can keep the config and the state between the calls. Does it make sense?

internal/logparse/logparse.go
0xgouda: We need to know the log dir in both cases, because even in remote parsing `pg_ls_logdir()` only returns the filenames, so we then do `logFilePath := filepath.Join(LogsDirPath, latestLogFile)` before using `pg_read_file($1, ...)`.

0xgouda:
> Instead of having three `ParseLogs` implementation, I'd really love to see

Good one. I also still have some TODOs here, such as implementing a graceful return on errors in local parsing. However, let's focus on getting the functionality working first. In another PR, we can refactor and make the code cleaner.

0xgouda:
> about naming. if we really want a separate package it should be `logparser`

renamed

pashagolub:
> In another PR, we can refactor and make the code cleaner.

Nope! :-) There is certainly no need to push anything to master if we can prettify it. Let's make struct with functions. And everything else can go afterwards

PR #1086, opened 2025-12-19
webui: fix preset sorting order and ESLint issues

Mahi1801:
Hi @0xgouda, thanks for the feedback!

I’ve moved the changes to a dedicated branch as suggested and updated the preset
sorting order to:

minimal
basic
standard
exhaustive
full
aiven
azure
gce
rds
pgbouncer
pgpool
unprivileged
recommendations
prometheus-async
exhaustive_no_python
debug

The same ordering is now used as the default in the Presets tab as well.
All checks are passing locally.

Please let me know if anything else is needed 🙏

### DIFF
--- a/internal/webui/src/constants/presets.ts
+++ b/internal/webui/src/constants/presets.ts
@@ -1,0 +1,18 @@
+export const PRESET_ORDER = [
+  "minimal",
+  "basic",
+  "standard",
+  "exhaustive",
+  "full",
+  "aiven",
+  "azure",
+  "gce",
+  "rds",
+  "pgbouncer",
+  "pgpool",
+  "unprivileged",
+  "recommendations",
+  "prometheus-async",
+  "exhaustive_no_python",
+  "debug",
+];
--- a/internal/webui/src/containers/SourceFormDialog/components/SourceForm/components/SourceFormStepMetrics.tsx
+++ b/internal/webui/src/containers/SourceFormDialog/components/SourceForm/components/SourceFormStepMetrics.tsx
@@ -1,3 +1,4 @@
+import { PRESET_ORDER } from "constants/presets";
 import { useEffect, useMemo } from "react";
 import DeleteIcon from "@mui/icons-material/Delete";
 import { Button, Checkbox, FormControl, FormControlLabel, FormHelperText, IconButton, InputLabel, OutlinedInput } from "@mui/material";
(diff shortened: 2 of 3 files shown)


0xgouda:
Thanks! The preset ordering in the sources form works as expected, but let's make this the default ordering in the presets tab too.

As you can see, they are not ordered by default:
image

pashagolub:
Thanks for your input! 

I don't think name list can cover all the situations. What if one wants to rename the preset? What if one adds their own preset and want to put it to the top? What if one is not happy with some preset and want to move it down? What if we delete some preset one day, e.g. `unprivileged`?

Should preset have some `order float` property that can be controlled via UI, so users may apply their own preferences?

Mahi1801:
Thanks for the thoughtful feedback!

I agree that a static name-based ordering cannot cover all future scenarios
like renaming presets, user-defined presets, or custom prioritization.
A configurable ordering (e.g. an explicit order/weight field controlled via UI)
would definitely be a more flexible long-term solution.

For this PR, I intentionally kept the scope limited to:
- fixing the current inconsistent ordering
- matching the expected default order across the source form and presets tab

This aligns with the existing backend data as it is today.
I’d be happy to explore or help with a follow-up improvement
if the project decides to support user-controlled preset ordering.

Please let me know how you’d like to proceed.

0xgouda:
please take a look at https://github.com/cybertec-postgresql/pgwatch/pull/1086#issuecomment-3674723329

Mahi1801:
Okay i got it. I will work on that.

Mahi1801:
Thanks for the clarification!

I've now applied the same predefined preset ordering as the default ordering
in the Presets grid as well. Presets not present in the predefined list are
placed after them.

Please let me know if you'd like this to be configurable in the future.

0xgouda:
> I've now applied the same predefined preset ordering as the default ordering in the Presets grid as well.

You haven't pushed any updates; it's still the same.

Mahi1801:
Hi @0xgouda, thanks for pointing that out -my apologies.
I’ve now pushed the update that applies the same predefined preset ordering to the Presets grid.
Please let me know if it looks correct now.

0xgouda:
LGTM, but please check the reviews from copilot.

Mahi1801:
Thanks for the review!

I’ve addressed the Copilot feedback by extracting the preset ordering into a shared constant and reusing it in both the Presets grid and Source form, removing duplication and improving consistency.

Formatting issues were also cleaned up.
Please let me know if everything looks good now

internal/webui/src/constants/presets.ts
0xgouda, line 1: Why add a new `constants/` folder and not use the already existing `consts/` folder?
(I am not an experienced front-end dev, correct me if I am missing smth)

internal/webui/src/constants/presets.ts
```diff
@@ -0,0 +1,18 @@
...
+  "gce",
+  "rds",
+  "pgbouncer",
+  "pgpool",
+  "unprivileged",
+  "recommendations",
+  "prometheus-async",
+  "exhaustive_no_python",
+  "debug",
+];
```
pashagolub, line 18: Front-end should not rely on the predefined preset list. We make assumption that no information is available at design-time. All information is available only through REST API

PR #1088, opened 2025-12-19
[-] fix `DeleteOldPartitions()` scheduling interval, fixes #1087

### LINKED ISSUE #1087: Expire old data every maintenance interval
Currently, old data is expired only every *retention interval*:
```
pgw.scheduleJob(pgw.retentionInterval, pgw.DeleteOldPartitions)
```

... which means that it's effectively keeping twice the amount of data.

It would make sense to switch to running that every *maintenance interval*. Especially useful when the retention interval is reduced and people are waiting for disk space to be freed.

I didn't check how often maintenance is performed. If it's not already the case, it should probably run right after startup so disk space gets freed when people need it.

0xgouda:
- Run `DeleteOldPartitions()` every `--maintenance-interval`, thus ensuring it runs regularly, removing expired data earlier.

Closes: #1087

### DIFF
--- a/internal/sinks/postgres.go
+++ b/internal/sinks/postgres.go
@@ -102,8 +102,10 @@ func NewWriterFromPostgresConn(ctx context.Context, conn db.PgxPoolIface, opts *
 		return
 	}
 
-	pgw.scheduleJob(pgw.maintenanceInterval, pgw.MaintainUniqueSources)
-	pgw.scheduleJob(pgw.maintenanceInterval, pgw.DeleteOldPartitions)
+	pgw.scheduleJob(pgw.maintenanceInterval, func() {
+		pgw.DeleteOldPartitions()
+		pgw.MaintainUniqueSources()
+	})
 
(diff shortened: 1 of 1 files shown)


internal/sinks/postgres.go
```diff
@@ -103,7 +103,7 @@ func NewWriterFromPostgresConn(ctx context.Context, conn db.PgxPoolIface, opts *
 	}
 
 	pgw.scheduleJob(pgw.maintenanceInterval, pgw.MaintainUniqueSources)
-	pgw.scheduleJob(pgw.retentionInterval, pgw.DeleteOldPartitions)
+	pgw.scheduleJob(pgw.maintenanceInterval, pgw.DeleteOldPartitions)
```
0xgouda: I think running the 2 routines with the same interval may create a lock contention on the sink DB? 

also we might want to ensure that `MaintainUniqueSources()` runs after `DeleteOldPartitions()`?

PR #1090, opened 2025-12-22
[-] apply source config updates to existing gatherers

### LINKED ISSUE #1084: Updates to metrics fetching intervals are not applied
**Describe the bug**
If using individual metrics updates to fetching intervals done via the webui or in the yaml files are never applied, in contrast, if using preset updates to fetching intervals in the preset definition are directly applied after a new metrics refresh.

**To Reproduce**
1. run pgwatch with individual metric (e.g., `wal: 10`)
2. Let it run for a while
3. update metric fetching interval to `60`
4. wait and observe that it will never be applied
5. run pgwatch with preset (e.g., `basic`)
6. Let it run for a while
7. update fetching intervals in the preset definition
8. Observe that it will get applied directly after sources + metrics refresh

**Expected behavior**
Updates to individual metrics fetching intervals must be applied.

### LINKED ISSUE #1085: Updates to connection strings are never applied to existing gatherers
**Describe the bug**
Any update to a source's connection string (e.g., param change, pointing to a different database, etc.) is never applied to already started gatherers

**To Reproduce**
1. Update `reaper.go:506` to show the conn str in the logs	`l.WithField("cache", fromCache).WithField("rows", len(data)).WithField("conn_str", md.ConnStr).Info("measurements fetched")`
2. Run pgwatch with a single source and some metrics
3. Update the conn str of the used source
4. Observe that the old gatherers will keep using the old conn str

**Expected behavior**
Should we allow updating the conn str of a source? If we should, this is a bug; if we shouldn't, we should disallow it from the web UI at least and mention it in the documentation.

**Screenshots**
two metrics for a single source fetching from 2 different databases because one of them is still using the old conn str:
```
2025-12-18 16:40:37.626 [INFO] [source:test] [metric:db_size] [rows:1] [cache:false] [conn_str:postgresql://postgres:postgres@localhost:5437/postgres] measurements fetched
2025-12-18 16:40:42.603 [INFO] [source:test] [metric:wal] [rows:1] [cache:true] [conn_str:postgresql://postgres:postgres@localhost:5437/pgwatch_metrics] measurements fetched
```

### LINKED ISSUE #1091: Updates to preset intervals are not applied in some scenarios
**Describe the bug**
In some scenarios (source config update), updates to a preset's metric fetching interval are not applied to running gatherers

**To Reproduce**
1. Run pgwatch with some preset, e.g., `basic`
2. Update the fetching interval for any metric in `basic`
3. Update the monitored source and add a standby preset.
4. Observe that the update to the fetching interval will never get applied

**Expected behavior**
Updates to metric fetching intervals should be applied.

**Additional context**
This happens because the update that added the standby preset, although not used anywhere, will cause `md.Equal()` in:

https://github.com/cybertec-postgresql/pgwatch/blob/4bae857896ec068dabeae7d4746fda5e9267eb63/internal/reaper/reaper.go#L398-L404

 to be false, hence dropping the old `*SourceConn` of that source from `r.monitoredSources` and use a new one, but the running gatherer will keep looking for updates from the old `*SourceConn` which has been dropped and will never get updated.

0xgouda:
- Detect metrics that their interval or conn str got changed and stop them, on the next reaper iteration they new gatherers with the latest configs will be automatically started

Closes #1084, Closes #1085, Closes #1091

### DIFF
--- a/internal/reaper/reaper.go
+++ b/internal/reaper/reaper.go
@@ -268,7 +268,7 @@ func (r *Reaper) ShutdownOldWorkers(ctx context.Context, hostsToShutDownDueToRol
 				if oldMd != nil {
 					oldInterval, ok := oldMd.Metrics[metric]
 					configChanged = ok &&
-						(oldInterval != currentMetricConfig[metric] ||
+						(oldInterval != currentMetricConfig[metric] && oldInterval > 0 ||
 							oldMd.ConnStr != md.ConnStr)
 				}
 			}


0xgouda:
Taken by #1106

PR #1097, opened 2025-12-27
UI: improve SQLs input with versioned SQL fields

### LINKED ISSUE #1092: Improvement for new metric form in UI
The current `SQLs` text box in the webui form that adds a new metric is not very intuitive. (Metrics tab => +New => SQLs)
With this box its hard to guess that its required to enter a list of "min version num": "sql" pairs. 
Image

Have 2 individual boxs one for postgres version number and the other for sql, with another button to add `New SQL`, mostly the same as the form for adding individual metrics under the sources tab:

Image
 

**Describe alternatives you've considered**
.

Ebraam-Ashraf:
### Problem
The current **SQLs** textarea in **Metrics → + New** is not intuitive.
Users have to guess that it expects a list of `"min_pg_version": "sql"` pairs,
which makes it easy to enter invalid or incorrect data.

### Solution
Replace the free-text SQLs textarea with a structured UI:
- Separate input for **Postgres minimum version**
- Separate textarea for **SQL**
- Ability to add multiple version → SQL pairs
- UI serializes the inputs back into the same payload format expected by the backend

### Validation
- At least **one version → SQL pair** is required before submission
- **Duplicate Postgres versions are not allowed**
- **Empty version or SQL fields** are rejected
- Validation is handled in the UI to prevent invalid payloads from being sent

Screenshot from 2025-12-27 13-53-20

Fixes #1092

### DIFF
--- a/cmd/pgwatch/doc.go
+++ b/cmd/pgwatch/doc.go
@@ -54,9 +54,13 @@
 //	                                     may be ignored by some sinks
 //	                                     (default: 950ms)
 //	                                     [$PW_BATCHING_DELAY]
-//	--retention=                         If set, metrics older than that will
-//	                                     be deleted (default: "14 days")
-//	                                     [$PW_RETENTION]
+//	--retention=                         Delete metrics older than this. 
+//	                                     Must be a valid PostgreSQL interval. 
+//	                                     (default: 14 days) [$PW_RETENTION]
+// --maintenance-interval=               Run pgwatch maintenance tasks on sinks with this interval 
+//										 e.g., deleting old metrics; Set to zero to disable. 
+//										 Must be a valid PostgreSQL interval. 
+// 										 (default: 12 hours) [$PW_MAINTENANCE_INTERVAL]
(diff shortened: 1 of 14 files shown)


pashagolub:
This approach wastes to much space
image

It might be a dropped down list with versions and one memo for SQL.

Ebraam-Ashraf:
@pashagolub 
thanks for feedback 

I understand you want the dropdown approach instead of separate entries
but if we use a dropdown with hardcoded versions (11, 12, 13...), won't we need to update the code with every new  release? 
and at some point, we'll need to remove old unsupported versions too.

also the UI layout where should the version dropdown be placed?
at the top, above the sql textarea?
or let it at its place ?

let me know your thoughts and I'll get it done

pashagolub:
> but if we use a dropdown with hardcoded versions (11, 12, 13...), won't we need to update the code with every new release?

dropdown can be editable. People can add and remove items from it. Maybe it should be tabs instead of dropdown. But accordion won't work. One SQL is sometimes 2-3 screens, so it just doesn't make sense

PR #1106, opened 2026-01-02
[-] fix source config update handling

0xgouda:
- Ensure that all gatherers have up-to-date configs (e.g., connections, intervals, etc.) from their source by restarting all gatherers under a specific source on config update.

Closes: #1085 
Closes: #1091 
Closes: #1084 
Closes: #1113

### DIFF
--- a/grafana/postgres/v12/0-health-check.json
+++ b/grafana/postgres/v12/0-health-check.json
@@ -5115,30 +5115,77 @@
       "type": "stat"
     },
     {
+      "datasource": {
+        "type": "grafana-postgresql-datasource",
+        "uid": "pgwatch-metrics"
+      },
+      "description": "Total size of user indexes for the selected DB. requires db_size metric.",
       "fieldConfig": {
-        "defaults": {},
+        "defaults": {
+          "decimals": 1,
+          "mappings": [],
+          "thresholds": {},
+          "unit": "bytes"
+        },
         "overrides": []
       },
       "gridPos": {
         "h": 2,
-        "w": 12,
+        "w": 6,
         "x": 12,
         "y": 18
       },
-      "id": 48,
+      "id": 51,
+      "maxDataPoints": 100,
       "options": {
-        "code": {
-          "language": "plaintext",
-          "showLineNumbers": false,
-          "showMiniMap": false
+        "colorMode": "none",
+        "graphMode": "none",
+        "justifyMode": "auto",
+        "orientation": "horizontal",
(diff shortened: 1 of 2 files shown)


internal/reaper/reaper.go
```diff
@@ -408,6 +408,21 @@ func (r *Reaper) LoadSources() (err error) {
...
+		} else {
+			currentMetricConfig = md.Metrics
+		}
+		for metric := range currentMetricConfig {
+			r.logger.WithField("source", md.Name).WithField("metric", metric).Info("stopping gatherer...")
+			dbMetric := md.Name + dbMetricJoinStr + metric
+			cancelFunc := r.cancelFuncs[dbMetric]
+			cancelFunc()
+			delete(r.cancelFuncs, dbMetric)
+		}
```
0xgouda: I used this instead of directly calling `r.ShutdownOldWorkers()` to avoid having a delete op syncmetric triggered.

but noting that starting the new gatherer with the updated configs will eventually trigger an add op syncmetric.

How do you see that?

internal/reaper/reaper.go
pashagolub: I'm confused now. We have at least 3 places where we're trying to catch and eliminate old workers. One is here.

pashagolub, line 179: And here is another place where we're trying to catch disabled metrics.

pashagolub:
We should consolidate the stopping functionality into one place and just call the routine whenever we think it's appropriate. What do you think @0xgouda?

0xgouda:
> We should consolidate the stopping functionality into one place and just call the routine whenever we think it's appropriate. What do you think @0xgouda?

I marked this as a draft because I was working on it; all will be using `ShutdownOldWorkers()`. Please re-check. I just pushed the changes.

pashagolub:
Oh, I see. I was confused by "review equested"

internal/reaper/database.go
```diff
@@ -527,18 +527,18 @@ func TryCreateMetricsFetchingHelpers(ctx context.Context, md *sources.SourceConn
 	return
 }
 
-func (r *Reaper) CloseResourcesForRemovedMonitoredDBs(shutDownDueToRoleChange map[string]bool) {
```
pashagolub, line 530: Good name! Old one is confusing as hell!

pashagolub:
Looks good. We might later improve cleaning routines.

pashagolub (approved):

PR #1115, opened 2026-01-07
[+] add `Total index size` panel to `Health Check` dashboard

0xgouda:
- Add `indexes_size_b` column to `db_size` metric
    - The column returns the total index size aggregated from `pg_stat_user_indexes`
- Add new `Total index size` panel to `Health Check` dashboard
---
- This is the first PR in a series of PRs that aim to improve the index dashboards
- I think the `Health Check` dashboard panels need to be re-arranged in a more logical grouping row-by-row, putting relevant panels together, but this is to be done in a separate PR

Screenshots:
image

---

image

### DIFF
--- a/internal/metrics/metrics.yaml
+++ b/internal/metrics/metrics.yaml
@@ -460,8 +460,8 @@ metrics:
                    from pg_class c join pg_namespace n on n.oid = c.relnamespace
                    where nspname = 'pg_catalog' and relkind = 'r'
                   ) as catalog_size_b,
-                  sum(pg_relation_size(indexrelid))::int8 as indexes_size_b
-                from pg_stat_user_indexes
+                  (select sum(pg_relation_size(indexrelid))::int8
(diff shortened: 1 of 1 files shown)


internal/metrics/metrics.yaml
```diff
@@ -458,7 +459,9 @@ metrics:
                   (select sum(pg_total_relation_size(c.oid))::int8
                    from pg_class c join pg_namespace n on n.oid = c.relnamespace
                    where nspname = 'pg_catalog' and relkind = 'r'
-                  ) as catalog_size_b
+                  ) as catalog_size_b,
+                  sum(pg_relation_size(indexrelid))::int8 as indexes_size_b
+                from pg_stat_user_indexes
```
pashagolub:
```suggestion
                  (select sum(pg_relation_size(indexrelid))::int8
                from pg_stat_user_indexes) as indexes_size_b
```
Should it be like this just to keep styling?

PR #1117, opened 2026-01-09
[*] refactor `reco_add_index` metric

0xgouda:
1. Refactor `reco_add_index` metric, returning each affected query in a separate row with the index ddl in the form: `query_id | recommended_index_ddl | ...`.
    
2. Order rows from step 1 by `total_exec_time` from `pg_stat_statements` and return the top 10 results.
    - Hence, returning the suggestions that optimize the most resource-intensive queries.
    
3. Add `Missing Indexes` panel to [`index overview`](https://demo.pgwatch.com/d/index-overview/index-overview?orgId=1&from=now-1h&to=now&timezone=browser&var-dbname=$__all) dashboard.
image

### DIFF
--- a/grafana/postgres/v12/index-overview.json
+++ b/grafana/postgres/v12/index-overview.json
@@ -26,7 +26,7 @@
         "type": "grafana-postgresql-datasource",
         "uid": "pgwatch-metrics"
       },
-      "description": "Adding indexes can have unexpected performance effects, so ensure that you test them first to verify that the recommendation has the desired effect.\n\nRequires reco_add_index metric. \n\nColumn Guide:\n- **Affected Queries**: number of queries affected by index creation.\n- **Heaviest Query**: the query with the biggest total execution time among all affected queries.",
+      "description": "Adding indexes can have unexpected performance effects, so ensure that you test them first to verify that the recommendation has the desired effect.\n\nRequires reco_add_index metric. \n\nColumn Guide:\n- **Affected Queries**: number of queries affected by index creation.\n- **Heaviest Query**: the query with the biggest total execution time among all affected queries.\n- **Recommendation**: the recommended index ddl.",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -618,6 +618,10 @@
               {
                 "id": "custom.align",
                 "value": "left"
+              },
+              {
+                "id": "decimals",
+                "value": 0
               }
             ]
           }


internal/metrics/metrics.yaml
```diff
@@ -1496,33 +1496,56 @@ metrics:
...
+                limit 10;
+            13: |-
+                with qrs as (
+                    select 
+                        (regexp_matches(indexes->>'ddl', E'ON (.*?) '))[1] as object_name,
+                        query_id::int8, 
+                        json_array_length(indexes->'queryids') as affected_query_count,
+                        indexes->>'ddl' as recommendation
+                    from 
+                        pg_qualstats_index_advisor(min_filter => 3000, min_selectivity => 85) pqa,
```
0xgouda: After diving into [pg_qualstats](https://github.com/powa-team/pg_qualstats/), I saw that we shouldn't rely on the default `min_selectivity = 30` and use higher numbers that filter a bigger percentage from the data, for better suggestion quality.

I decided to use `min_filter => 3000` and `min_selecitivity => 85%`, which only suggests indexes on predicates that filter at least 85% of the data and requires a minimum of 3000 rows filtered. Please let me know what you think about those numbers.

internal/metrics/metrics.yaml
```diff
@@ -1496,33 +1496,56 @@ metrics:
...
+                limit 10;
+            13: |-
+                with qrs as (
+                    select 
+                        (regexp_matches(indexes->>'ddl', E'ON (.*?) '))[1] as object_name,
+                        query_id::int8, 
+                        json_array_length(indexes->'queryids') as affected_query_count,
+                        indexes->>'ddl' as recommendation
+                    from 
+                        pg_qualstats_index_advisor(min_filter => 3000, min_selectivity => 90) pqa,
```
0xgouda, line 1531: After diving into [pg_qualstats](https://github.com/powa-team/pg_qualstats/), I saw that we shouldn't rely on the default `min_selectivity = 30` and use higher numbers that filter a bigger percentage from the data, for better suggestion quality.

I decided to use `min_filter => 3000` and `min_selecitivity => 90%`, which only suggests indexes on predicates that filter at least 90% of the data and requires a minimum of 3000 rows filtered. Please let me know what you think about those numbers.

pashagolub: sounds good!

internal/metrics/metrics.yaml
```diff
@@ -1496,33 +1496,56 @@ metrics:
...
-        init_sql: "CREATE EXTENSION IF NOT EXISTS pg_qualstats;"
+                    qrs.query_id,
+                    qrs.affected_query_count,
+                    qrs.recommendation
+                from 
+                    qrs
+                join pg_stat_statements s on qrs.query_id = s.queryid
+                where object_name > '' and recommendation > ''
+                order by object_name, qrs.recommendation, s.total_time desc
+                limit 10;
```
pashagolub, line 1522: we could drop pre-14 support, so it's OK to remove 11 version sql

pashagolub (approved):

internal/metrics/metrics.yaml
0xgouda: will get rid of all metric sqls for pre-14 versions, in a separate PR.

PR #1121, opened 2026-01-15
[*] refactor `reco_drop_index` metric

### LINKED ISSUE #1118: Improvement for `index stats` metrics and dashboards
While working on #1117, I noticed that the `Duplicate Indexes`, `Invalid Indexes`, and `Unused Indexes` panels in the [Index Overview](https://demo.pgwatch.com/d/index-overview/index-overview?orgId=1&from=now-1h&to=now&timezone=browser&var-dbname=$__all) dashboard rely on the `index_stats` metric, while the `reco_drop_index` metric also measures these same factors, but is not used.

It would be better to refactor the `reco_drop_index` metric and rely on it instead of `index_stats`, as this will decrease the number of rows required from `index_stats`, which has a much lower fetching interval (15 minutes) compared to `reco_drop_index` (18 hours)

Hence, improving performance and storage.

0xgouda:
- Refactor `reco_drop_index` metric to measure duplicate indexes and return each index entry in a separate row
- Remove measuring `invalid/unused` indexes from `index_stats` its now the responsibility of `reco_drop_index`

Updated `index overview` dashboard view:

image

Prometheus `Recommendations` dashboard:

image

Closes #1118

### DIFF
--- a/docs/concept/components.md
+++ b/docs/concept/components.md
@@ -49,11 +49,9 @@ language:
 
 ### [PostgreSQL](https://www.postgresql.org/)
 
-PostgreSQL is a world's most advanced Open Source RDBMS.
+PostgreSQL is the world's most advanced Open Source RDBMS.
 
-Postgres storage is based on the JSONB datatype so minimally
-version 9.4+ is required, but for bigger setups where partitioning
-is a must, v11+ is needed. Any already existing Postgres database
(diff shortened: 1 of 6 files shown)


pashagolub:
#1123 causes conflicts here

pashagolub (approved):
LGTM! Thanks!

PR #1127, opened 2026-01-16
[-] only create time partitions if needed

### LINKED ISSUE #1126: pgwatch creates 4 new partitions each time it starts.
**Describe the bug**
pgwatch creates 4 new partitions each time it gets restarted even if those partitions are not needed (locally i have partitions up to July 2026)

**To Reproduce**
1. Run pgwatch
2. Stop pgwatch
3. Check the number of partitions for any metric table
4. Repeat and observe the change by 4 each time

**Expected behavior**
Partitions should be created only if needed, or at most 3 partitions in the future.

**Screenshots**
partitions for source, I just ran a few minutes ago, but had to restart many times:
```
 subpartitions."archiver_pending_count_omdb.test_20260112"         | 2026-01-19 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260119"         | 2026-01-26 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260126"         | 2026-02-02 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260202"         | 2026-02-09 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260209"         | 2026-02-16 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260216"         | 2026-02-23 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260223"         | 2026-03-02 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260302"         | 2026-03-09 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260309"         | 2026-03-16 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260316"         | 2026-03-23 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260323"         | 2026-03-30 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260330"         | 2026-04-06 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260406"         | 2026-04-13 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260413"         | 2026-04-20 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260420"         | 2026-04-27 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260427"         | 2026-05-04 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260504"         | 2026-05-11 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260511"         | 2026-05-18 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260518"         | 2026-05-25 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260525"         | 2026-06-01 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260601"         | 2026-06-08 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260608"         | 2026-06-15 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260615"         | 2026-06-22 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260622"         | 2026-06-29 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260629"         | 2026-07-06 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260706"         | 2026-07-13 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260713"         | 2026-07-20 00:00:00
 subpartitions."archiver_pending_count_omdb.test_20260720"         | 2026-07-27 00:00:00
```

0xgouda:
- Check if the needed time partition already exists before creating new partitions in postgres sink

TODO:
- [x] Add tests
- [ ] Ensure a migration exists for the changes

Closes #1126

### DIFF
--- a/internal/sinks/postgres.go
+++ b/internal/sinks/postgres.go
@@ -415,7 +415,7 @@ func (pgw *PostgresWriter) flush(msgs []metrics.MeasurementEnvelope) {
 
 	switch pgw.metricSchema {
 	case DbStorageSchemaPostgres:
-		err = pgw.EnsureMetricDbnameTime(pgPartBoundsDbName, pgw.forceRecreatePartitions)
+		err = pgw.EnsureMetricDbnameTime(pgPartBoundsDbName)
 	case DbStorageSchemaTimescale:
 		err = pgw.EnsureMetricTimescale(pgPartBounds)
 	default:
@@ -465,7 +465,7 @@ func (pgw *PostgresWriter) EnsureMetricTimescale(pgPartBounds map[string]Existin
 	return
 }
 
-func (pgw *PostgresWriter) EnsureMetricDbnameTime(metricDbnamePartBounds map[string]map[string]ExistingPartitionInfo, force bool) (err error) {
+func (pgw *PostgresWriter) EnsureMetricDbnameTime(metricDbnamePartBounds map[string]map[string]ExistingPartitionInfo) (err error) {
 	var rows pgx.Rows
 	sqlEnsure := `select * from admin.ensure_partition_metric_dbname_time($1, $2, $3, $4)`
 	for metric, dbnameTimestampMap := range metricDbnamePartBounds {
@@ -479,7 +479,7 @@ func (pgw *PostgresWriter) EnsureMetricDbnameTime(metricDbnamePartBounds map[str
 				return fmt.Errorf("zero StartTime/EndTime in partitioning request: [%s:%v]", metric, pb)
 			}
 			partInfo, ok := pgw.partitionMapMetricDbname[metric][dbname]
-			if !ok || (ok && (pb.StartTime.Before(partInfo.StartTime))) || force {
+			if !ok || (ok && (pb.StartTime.Before(partInfo.StartTime))) || pgw.forceRecreatePartitions {
 				if rows, err = pgw.sinkDb.Query(pgw.ctx, sqlEnsure, metric, dbname, pb.StartTime, pgw.opts.PartitionInterval); err != nil {
 					return
 				}
@@ -488,7 +488,7 @@ func (pgw *PostgresWriter) EnsureMetricDbnameTime(metricDbnamePartBounds map[str
 				}
 				pgw.partitionMapMetricDbname[metric][dbname] = partInfo
 			}
-			if pb.EndTime.After(partInfo.EndTime) || pb.EndTime.Equal(partInfo.EndTime) || force {
+			if pb.EndTime.After(partInfo.EndTime) || pb.EndTime.Equal(partInfo.EndTime) || pgw.forceRecreatePartitions {
 				if rows, err = pgw.sinkDb.Query(pgw.ctx, sqlEnsure, metric, dbname, pb.EndTime, pgw.opts.PartitionInterval); err != nil {
 					return
 				}
--- a/internal/sinks/postgres_test.go
+++ b/internal/sinks/postgres_test.go
@@ -694,7 +694,7 @@ func TestPartitionInterval(t *testing.T) {
 			},
 		},
 	}
-	err = pgw.EnsureMetricDbnameTime(m, false)
+	err = pgw.EnsureMetricDbnameTime(m)
 	r.NoError(err)
 
 	var partitionsNum int


internal/sinks/sql/ensure_partition_postgres.sql
```diff
@@ -119,6 +119,8 @@ BEGIN
           IF metric_timestamp >= l_part_start AND metric_timestamp < l_part_end THEN
               part_available_from := l_part_start;
               part_available_to := l_part_end;
+          ELSEIF metric_timestamp < l_part_start THEN
+              EXIT; -- No need to create further partitions
           END IF;
```
0xgouda, line 124: The added condition returns and avoids creating additional partitions if we already have the required partition and at least 1 more pre-created partition after it.

However, this now results in the assumption that the opposing `flush(msgs)` metrics won't have an upper bound timestamp > partitions_upper_bound, or this will cause a bug having no time partitions for these rows.

@pashagolub, what do you think? Can we trust this assumption?

0xgouda: If a user has `--partition-interval` >= `--batching-delay`, that might trigger this bug, but that's a very werid choice.

and btw the current implementation assumes that a single `flush(msgs)` should fit inside `8` partitions at most (the maximum number of partitions that would be created by a single call to `EnsureMetricDbnameTime()`)

pashagolub: flush() always checks if there are partiotions pre-created, AFAIR

```go
	switch pgw.metricSchema {
	case DbStorageSchemaPostgres:
		err = pgw.EnsureMetricDbnameTime(pgPartBoundsDbName, forceRecreatePartitions)
	case DbStorageSchemaTimescale:
		err = pgw.EnsureMetricTimescale(pgPartBounds)
```

So partiotions are created on demand

pashagolub: That said, it's safe to assume your statement. But we also may add test to cover that case and check :)

0xgouda: These lines are the main source of the issue. Please look at #1126, they create 4 new partitions each time pgwatch restarts, also i explained in the `Additional Context` section why they gradually cause more than 3 partitions to be pre-created even without a restart.

internal/sinks/postgres.go
```diff
@@ -202,6 +202,13 @@ var (
 	partitionMapMetricDbname = make(map[string]map[string]ExistingPartitionInfo) // metric[dbname = min/max bounds]
 )
 
+// ResetPartitionCache clears the partition cache 
+func ResetPartitionCache() {
+	forceRecreatePartitions = false
+	partitionMapMetric = make(map[string]ExistingPartitionInfo)
+	partitionMapMetricDbname = make(map[string]map[string]ExistingPartitionInfo)
+}
```
pashagolub: I see what you're doing here! :)
We don't want global functions in production code used only in tests. I understand why you did this. 
I believe it's time to move 
```go
forceRecreatePartitions 
partitionMapMetric      
partitionMapMetricDbname
```
inside `PostgresWriter`, so every time you re-create the writer, it will create partition-related things and nullify those.

pashagolub (changes requested):

internal/sinks/postgres.go
0xgouda: done, please re-review

internal/sinks/postgres.go
```diff
@@ -415,13 +415,13 @@ func (pgw *PostgresWriter) flush(msgs []metrics.MeasurementEnvelope) {
 
 	switch pgw.metricSchema {
 	case DbStorageSchemaPostgres:
-		err = pgw.EnsureMetricDbnameTime(pgPartBoundsDbName, forceRecreatePartitions)
+		err = pgw.EnsureMetricDbnameTime(pgPartBoundsDbName, pgw.forceRecreatePartitions)
```
pashagolub: should we omit the second parameter since it is a pgw field anyway, e.g.


```suggestion
		err = pgw.EnsureMetricDbnameTime(pgPartBoundsDbName)
```

PR #1128, opened 2026-01-17
[-] fix log parser offset reset on file switch, closes #1089

### LINKED ISSUE #1089: Problem in log parser logic
**Describe the bug**
The current log parser implementation starts reading from offset zero on opening new log file, with configs like `log_truncate_on_rotation = off` and `log_file_name = postgresql-%H`, the files will be reused and not truncated, hence pgwatch will read all of them from the beginning each time they become the latest modified file. 

**Expected behavior**
pgwatch should start from the offset it was it the last time it saw the file.

ijaidev:
## Fix log parser offset reset on file switch

Fixes #1089 

### Problem
When using PostgreSQL with hourly log rotation (`log_filename = postgresql-%H.csv`) and `log_truncate_on_rotation = off`, the log parser was resetting its read offset to 0 when switching back to previously-seen files. This caused duplicate log entries to be counted every time a file was reused (e.g., 24 hours later).

### Root Cause
In both `logparser_remote.go` (line 87) and `logparser_local.go` (line 111), the offset/line count was unconditionally reset to 0 when switching files:
```go
offset = 0  // BUG: Always resets, even for previously-seen files
```

### Solution
Added in-memory offset tracking per log file:
- Added `fileOffsets map[string]int64` to `LogParser` struct
- Save current file's offset before switching
- Restore offset when returning to a previously-seen file
- Added `pruneFileOffsets()` to clear map when exceeding 100 entries (prevents unbounded memory growth)
- Added unit tests for offset tracking and pruning logic

### Changes
| File | Change |
|------|--------|
| `logparser.go` | Added `fileOffsets` field and `pruneFileOffsets()` method |
| `logparser_remote.go` | Save/restore byte offsets on file switch |
| `logparser_local.go` | Save/restore line counts on file switch |
| `logparser_test.go` | Added `TestFileOffsetsTrackingAndPruning` |

### Testing
- [x] Docker build successful
- [x] pgwatch runs and parses logs correctly
- [x] Unit tests added for offset tracking and pruning

### DIFF
--- a/internal/reaper/logparser.go
+++ b/internal/reaper/logparser.go
@@ -35,18 +35,27 @@ const csvLogDefaultRegEx = `^^(?P<log_time>.*?),"?(?P<user_name>.*?)"?,"?(?P<dat
 const csvLogDefaultGlobSuffix = "*.csv"
 
 const maxChunkSize int32 = 10 * 1024 * 1024 // 10 MB
+const maxTrackedFiles = 100                 // max log files to track offsets for
+
+// fileState stores the parsing state for a log file
+type fileState struct {
+	offset   int64 // last read position (bytes for remote, lines for local)
+	fileSize int64 // file size when offset was saved (for truncation detection)
+}
 
 type LogParser struct {
-	ctx                context.Context
-	LogsMatchRegex     *regexp.Regexp
-	LogFolder          string
-	ServerMessagesLang string
-	SourceConn         *sources.SourceConn
-	Interval           float64
-	StoreCh            chan<- metrics.MeasurementEnvelope
-	eventCounts        map[string]int64 // for the specific DB. [WARNING: 34, ERROR: 10, ...], zeroed on storage send
-	eventCountsTotal   map[string]int64 // for the whole instance
-	lastSendTime       time.Time
+	ctx                   context.Context
+	LogsMatchRegex        *regexp.Regexp
+	LogFolder             string
+	ServerMessagesLang    string
+	SourceConn            *sources.SourceConn
+	Interval              float64
+	StoreCh               chan<- metrics.MeasurementEnvelope
+	eventCounts           map[string]int64 // for the specific DB. [WARNING: 34, ERROR: 10, ...], zeroed on storage send
+	eventCountsTotal      map[string]int64 // for the whole instance
+	lastSendTime          time.Time
+	fileStates            map[string]*fileState // tracks parsing state per log file
+	logTruncateOnRotation bool                  // PostgreSQL's log_truncate_on_rotation setting
 }
 
 func NewLogParser(ctx context.Context, mdb *sources.SourceConn, storeCh chan<- metrics.MeasurementEnvelope) (*LogParser, error) {
@@ -74,16 +83,25 @@ func NewLogParser(ctx context.Context, mdb *sources.SourceConn, storeCh chan<- m
 		return nil, err
 	}
 
+	var logTruncateOnRotation bool
+	if logTruncateOnRotation, err = tryDetermineLogTruncateOnRotation(ctx, mdb.Conn); err != nil {
+		logger.WithError(err).Warning("Could not determine log_truncate_on_rotation setting, assuming off")
+		logTruncateOnRotation = false
+	}
+	logger.Debugf("log_truncate_on_rotation is set to: %v", logTruncateOnRotation)
+
 	return &LogParser{
-		ctx:                ctx,
-		LogsMatchRegex:     logsRegex,
-		LogFolder:          logFolder,
-		ServerMessagesLang: serverMessagesLang,
-		SourceConn:         mdb,
-		Interval:           mdb.GetMetricInterval(specialMetricServerLogEventCounts),
-		StoreCh:            storeCh,
-		eventCounts:        make(map[string]int64),
-		eventCountsTotal:   make(map[string]int64),
+		ctx:                   ctx,
+		LogsMatchRegex:        logsRegex,
+		LogFolder:             logFolder,
+		ServerMessagesLang:    serverMessagesLang,
+		SourceConn:            mdb,
+		Interval:              mdb.GetMetricInterval(specialMetricServerLogEventCounts),
+		StoreCh:               storeCh,
+		eventCounts:           make(map[string]int64),
+		eventCountsTotal:      make(map[string]int64),
+		fileStates:            make(map[string]*fileState),
+		logTruncateOnRotation: logTruncateOnRotation,
 	}, nil
 }
 
@@ -134,6 +152,16 @@ func tryDetermineLogMessagesLanguage(ctx context.Context, conn db.PgxIface) (str
 		return "en", nil
 	}
 	return lc, nil
+}
+
+func tryDetermineLogTruncateOnRotation(ctx context.Context, conn db.PgxIface) (bool, error) {
+	sql := `select current_setting('log_truncate_on_rotation')`
+	var setting string
+	err := conn.QueryRow(ctx, sql).Scan(&setting)
+	if err != nil {
+		return false, err
+	}
+	return setting == "on", nil
 }
 
 func checkHasRemotePrivileges(ctx context.Context, mdb *sources.SourceConn, logsDirPath string) error {
--- a/internal/reaper/logparser_local.go
+++ b/internal/reaper/logparser_local.go
@@ -80,6 +80,7 @@ func (lp *LogParser) parseLogsLocal() error {
 				logger.Debugf("Skipped %d already processed lines from %s", linesRead, latest)
 			} else if firstRun { // seek to end
 				_, _ = latestHandle.Seek(0, 2)
+				reader.Reset(latestHandle)
 				firstRun = false
 			}
 		}
@@ -94,6 +95,23 @@ func (lp *LogParser) parseLogsLocal() error {
 			}
 
 			if err == io.EOF {
+				// Check for truncation of current file
+				if fi, err := latestHandle.Stat(); err == nil {
+					if sysOffset, err := latestHandle.Seek(0, 1); err == nil {
+						if sysOffset > fi.Size() {
+							logger.Infof("File %s truncated (size %d < offset %d). Resetting...", latest, fi.Size(), sysOffset)
+							if _, err := latestHandle.Seek(0, 0); err == nil {
+								reader.Reset(latestHandle)
+								linesRead = 0
+								continue
+							}
+						} else if sysOffset < fi.Size() {
+							// File grew but reader hit EOF. Reset reader to ensure it picks up new content.
+							reader.Reset(latestHandle)
+						}
+					}
+				}
+
 				// // EOF reached, wait for new files to be added
 				select {
 				case <-lp.ctx.Done():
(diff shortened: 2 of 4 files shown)


internal/reaper/logparser_remote.go
```diff
@@ -82,10 +82,19 @@ func (lp *LogParser) parseLogsRemote() error {
...
+						lp.fileOffsets[latestLogFile] = int64(offset)
+						lp.pruneFileOffsets() // prevent unbounded map growth
 						latestLogFile = fileName
 						size = latestSize
-						offset = 0
-						logger.Infof("Switching to new logfile: '%s'", fileName)
+						// Restore offset if we've seen this file before and it hasn't been truncated
+						if savedOffset, ok := lp.fileOffsets[fileName]; ok && savedOffset <= int64(latestSize) {
+							offset = int32(savedOffset)
+							logger.Infof("Switching to logfile: '%s' (resuming from offset %d)", fileName, offset)
```
0xgouda: What if the file has been truncated and the new offset is bigger than or equal to our saved offset?

internal/reaper/logparser_local.go
```diff
@@ -103,12 +103,21 @@ func (lp *LogParser) parseLogsLocal() error {
...
 					latest = file
 					_ = latestHandle.Close()
 					latestHandle = nil
-					logger.Infof("Switching to new logfile: %s", file)
-					linesRead = 0
+					// Restore line count if we've seen this file before
+					if savedLines, ok := lp.fileOffsets[file]; ok {
+						linesRead = int(savedLines)
+						logger.Infof("Switching to logfile: %s (resuming from line %d)", file, linesRead)
+					} else {
```
0xgouda: Same as above + why isn't file truncation considered at all here?

internal/reaper/logparser_test.go
```diff
@@ -349,6 +349,104 @@ func TestZeroEventCounts(t *testing.T) {
...
+
+		// Verify both offsets are saved
+		assert.Equal(t, int64(50000), lp.fileOffsets["postgresql-10.csv"])
+		assert.Equal(t, int64(75000), lp.fileOffsets["postgresql-11.csv"])
+
+		// Simulate restoring offset when switching back to file1
+		restoredOffset, ok := lp.fileOffsets["postgresql-10.csv"]
+		assert.True(t, ok, "should find saved offset")
+		assert.Equal(t, int64(50000), restoredOffset)
+	})
```
0xgouda: I don't see any useful stuff that this tests.

internal/reaper/logparser_test.go
```diff
@@ -349,6 +349,104 @@ func TestZeroEventCounts(t *testing.T) {
...
+		savedOffset, ok = lp.fileOffsets["postgresql-10.csv"]
+
+		if ok && savedOffset <= newFileSize {
+			// File not truncated, use saved offset
+			assert.Equal(t, int64(50000), savedOffset)
+		} else {
+			assert.Fail(t, "should use saved offset when file has grown")
+		}
+	})
+}
```
0xgouda: `assert.True(t, true, "correctly detected truncation")` 
and 
`assert.Fail(t, "should use saved offset when file has grown")` 
don't indicate anything.

also you shouldn't use the `ok && savedOffset <= newFileSize` operator itself; you should invoke the real function and see how it behaves. There is no function from the actual code called here.

internal/reaper/logparser_remote.go
ijaidev: I want to discuss about this 

We don't have an accurate method to check if the file is truncated in this case. We are assuming that `log_truncate_on_rotation = off` So the file will only grow.

One idea is in mind is to store some of the latest bytes of the file and compare them with the newly updated file.

What do you think?

0xgouda: you can use `select current_setting('log_truncate_on_rotation');` on parser instantiation to know whether it's on or off

ijaidev: Ohh, I missed it

If it's on
We are assuming that the file will only grow and there are no external delete or truncate operations, right?

0xgouda: > If it's on
> We are assuming that the file will only grow and there are no external delete or truncate operations, right?

If it's "off"

ijaidev: Yes, off. My bad

Thanks for the guidance, working on it

internal/reaper/logparser_remote.go
```diff
@@ -75,17 +75,41 @@ func (lp *LogParser) parseLogsRemote() error {
 				err := lp.SourceConn.Conn.QueryRow(lp.ctx, "select size, modification from pg_ls_logdir() where name = $1;", latestLogFile).Scan(&latestSize, &modification)
 				if err != nil {
 					logger.Warnf("Failed to read state info of logfile: '%s'", latestLogFile)
+				} else if latestSize < size {
+					logger.Infof("File %s truncated (size reduced from %d to %d). Resetting offset to 0.", latestLogFile, size, latestSize)
+					offset = 0
+					size = latestSize
+					linesRead = 0
+					numOfLines = 0
+					break
```
0xgouda, line 84: What is this block trying to do? 

The file will never be truncated unless its being newly opened (switching from another file), isn't it?

which is handled below.

internal/reaper/logparser_remote.go
```diff
@@ -75,17 +75,41 @@ func (lp *LogParser) parseLogsRemote() error {
...
 						size = latestSize
-						offset = 0
-						logger.Infof("Switching to new logfile: '%s'", fileName)
+						// Restore offset if we've seen this file before and it hasn't been truncated
+						// Only restore offset if:
+						// 1. We have a saved state for this file
+						// 2. log_truncate_on_rotation is OFF (file only grows)
+						// 3. The current file size >= saved file size (not truncated)
+						if state, ok := lp.fileStates[fileName]; ok && !lp.logTruncateOnRotation && int64(latestSize) >= state.fileSize {
+							offset = int32(state.offset)
```
0xgouda: Why are we saving the file size then rechecking that `latestSize >= state.fileSize`?

I don't get why not save the offset only and then start from there?

internal/reaper/logparser.go
```diff
@@ -35,18 +35,27 @@ const csvLogDefaultRegEx = `^^(?P<log_time>.*?),"?(?P<user_name>.*?)"?,"?(?P<dat
 const csvLogDefaultGlobSuffix = "*.csv"
 
 const maxChunkSize int32 = 10 * 1024 * 1024 // 10 MB
+const maxTrackedFiles = 100                 // max log files to track offsets for
+
+// fileState stores the parsing state for a log file
+type fileState struct {
+	offset   int64 // last read position (bytes for remote, lines for local)
+	fileSize int64 // file size when offset was saved (for truncation detection)
+}
```
0xgouda, line 44: Same as above, why do we need a new `fileState` struct? do we need the file size?

0xgouda:
Thanks for taking the time to refactor!

New loggings (and comments) are a bit verbose + `pruneFileOffsets()` logic doesn't need to be in a separate function invokding the map cleaning code directly should be enough and more readable.

internal/reaper/logparser_remote.go
ijaidev: You are right for setups that use date patterns in the filename (e.g., postgresql-%a.log), because rotation forces a name change (Sun -> Mon).

However, if a user configures a static filename (e.g., log_filename = 'postgresql.csv') and uses log_truncate_on_rotation, the filename does not change during rotation, only the size drops. (or file size drops for any other reason)

Without this block, the parser sees latestLogFile == fileName, skips the reset logic below, and crashes trying to read the old offset on the truncated file. This block is necessary to support static filename configurations.

it's a safety check

internal/reaper/logparser_remote.go
ijaidev: For remote, we can simplify to just saving offset and checking latestSize < offset. For local parsing, we use file size (bytes) instead of line count because getting exact line count requires iterating through the file, while os.Stat().Size() is O(1).

to maintain consistency, I used the same approach in both parsers.

internal/reaper/logparser.go
ijaidev: I explained above :)

ijaidev:
Hey, Thanks for the review

I have removed the verbose comments and loggings

and made the cleaning logic inline 

Have a look

internal/reaper/logparser_remote.go
0xgouda: > However, if a user configures a static filename (e.g., log_filename = 'postgresql.csv') and uses log_truncate_on_rotation, the filename does not change during rotation, only the size drops. (or file size drops for any other reason)

have you reproduced that? because iirc the file won't be truncated unless its a new file (with a new name) being opened.

internal/reaper/logparser_remote.go
0xgouda: thanks for you time!

If fileSize is not required we shouldn't use it.

ijaidev: file size is required in the local parser. I can use the offset in the remote parser if you want.

0xgouda: ok plase update the remote parser, I will take a look at the local parser tomorrow or so.

internal/reaper/logparser_remote.go
ijaidev: You're right

 I reproduced this with Docker and confirmed PostgreSQL does NOT truncate a static log file in-place. 
 
The `log_truncate_on_rotation` setting only applies when the filename changes during rotation. 

I can remove this block if you prefer, or keep it as a safety check for external rotation tools (like logrotate copytruncate).

0xgouda: Yeah, please remove it.

internal/reaper/logparser_local.go
```diff
@@ -94,6 +95,23 @@ func (lp *LogParser) parseLogsLocal() error {
...
+								linesRead = 0
+								continue
+							}
+						} else if sysOffset < fi.Size() {
+							// File grew but reader hit EOF. Reset reader to ensure it picks up new content.
+							reader.Reset(latestHandle)
+						}
+					}
+				}
+
```
0xgouda, line 114: I don't get what this block is trying to do?

internal/reaper/logparser_local.go
```diff
@@ -103,12 +121,30 @@ func (lp *LogParser) parseLogsLocal() error {
...
+					// Restore offset if file wasn't truncated
+					if state, ok := lp.fileStates[file]; ok && !lp.logTruncateOnRotation {
+						if fi, err := os.Stat(file); err == nil && fi.Size() >= state.fileSize {
+							linesRead = int(state.offset)
+							logger.Infof("Resuming logfile %s from line %d", file, linesRead)
+						} else {
+							linesRead = 0
+							logger.Infof("Logfile %s was truncated, starting from beginning", file)
+						}
+					} else {
```
0xgouda, line 144: Why there is 2 conditions? if its not truncated then we guarantee that the size will be `>=` the last time we checked it?

0xgouda:
Please note that the issue is already simple, I don't know why we had to spend that much time on it and go through several reviews, please take the simplest approach for solving things!

pashagolub:
I close the PR as non-productive! Thanks for your contribution!

ijaidev:
Hey, Thanks for the review 

Sorry, didn't work :) I would have tried a simpler approach with less verbose. I'm new to this codebase.

I would like to propose an alternative PR with a simpler approach. Should I?

0xgouda:
Hi @ijaidev, it has been resolved in #1168.

Sorry, we had to finalize it as we are probably releasing v5-beta tomorrow.

Thanks for your time, and you are welcome to continue contributing!

PR #1130, opened 2026-01-17
[*] improve Prometheus Health-check dashboard

kmoppel:
Plus convert Max. connections into more useful % of total connection slots used.

Rename "Active connections" to more clear "Active queries"

### DIFF
--- a/grafana/postgres/v12/0-health-check.json
+++ b/grafana/postgres/v12/0-health-check.json
@@ -19,7 +19,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 2,
+  "id": 17,
   "links": [],
   "panels": [
     {
@@ -554,7 +554,7 @@
         "type": "grafana-postgresql-datasource",
         "uid": "pgwatch-metrics"
       },
-      "description": "Connected sessions, including Postgres internal processes like Autovacuum. < 1 Warn / > 300 Error thresholds by default",
+      "description": "",
       "fieldConfig": {
         "defaults": {
           "decimals": 0,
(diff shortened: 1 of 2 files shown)


kmoppel:
Screenshot from 2026-01-18 00-38-21

pashagolub:
Thanks!

@0xgouda would you please sync with the pg dashboard? Thanks

0xgouda:
> Thanks!
> 
> @0xgouda would you please sync with the pg dashboard? Thanks

Wal archiving status already exists in pg dashboards.
Update `Max connections` to `Connections used %`:
image

However, one thing to note: 
Active queries/connections in Prometheus dashboards queries `backends` metric and display the client connections count, whereas Active connections in the pg dashboard queries' db_stats' for all connected backends, including internal processes.

They should both query the same metric, but I wasn't sure which one to change, what do yout think?

pashagolub:
@kmoppel what do you think? We need to sync both dashboards

kmoppel:
Ok when need to sync: I'd myself prefer the Prom equivalent - active queries based on `backends.active`, as the new "Connections used %" already kind of includes an indication of total connections :)

kmoppel:
PS full syncing I think is not 100% possible anyways - as some metrics are missing and there are quite some complex SQL based panels that can't be even replicated with PromQL

0xgouda:
> Ok when need to sync: I'd myself prefer the Prom equivalent - active queries based on `backends.active`, as the new "Connections used %" already kind of includes an indication of total connections :)

Ok, updated accordingly, but a noting that the `Connections used %` were also querying `backends` for active client connections i moved it to query `db_stats` to get the total clients in both prom and pg dashboards.

image

image

PR #1132, opened 2026-01-19
Fix sink upgrade - error to warning

### LINKED ISSUE #1114: Can't upgrade sink schema if using yaml source and metric configs
**Describe the bug**

**To Reproduce**

**Expected behavior**

Mohamed-Elfardy:
Allow sink-only config upgrades when sources or metrics storage do not support upgrades.

Unsupported upgrades now emit a warning instead of failing, and the sink schema upgrade proceeds as expected.
The "--version" output was also updated to include the sink schema version.
the error message turned into warning.

Fixes #1114

### DIFF
--- a/cmd/pgwatch/version.go
+++ b/cmd/pgwatch/version.go
@@ -7,8 +7,8 @@ var (
 	commit       = "unknown"
 	version      = "unknown"
 	date         = "unknown"
-	ConfigSchema = "00824"
-	SinkSchema   = "01110"
+	configSchema = "00824"
+	sinkSchema   = "01110"
 )
 
 func printVersion() {
@@ -20,5 +20,5 @@ Version info:
   Git Commit:    %s
   Built:         %s
 
-`, version, ConfigSchema, SinkSchema, commit, date)
+`, version, configSchema, sinkSchema, commit, date)
 }
--- a/internal/cmdopts/cmdconfig.go
+++ b/internal/cmdopts/cmdconfig.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 
-	"github.com/cybertec-postgresql/pgwatch/v5/internal/log"
 	"github.com/cybertec-postgresql/pgwatch/v5/internal/metrics"
 	"github.com/cybertec-postgresql/pgwatch/v5/internal/sources"
 )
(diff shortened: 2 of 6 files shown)


0xgouda:
please fix the failing linter test.

Mohamed-Elfardy:
> please fix the failing linter test.

Hi,
No code changes here — this is just an empty commit to re-run CI after the previous workflow failure.

The workflow is waiting for your approval.
Thanks!

README.md
```diff
@@ -1,3 +1,4 @@
+Minor fix for sink upgrade warning
```
pashagolub: Why this line is here?

Mohamed-Elfardy: This line was added unintentionally during a local issue I was debugging.
It’s not needed and I’ll remove it. Thanks for the review.

cmd/pgwatch/version.go
```diff
@@ -4,19 +4,21 @@ import "fmt"
...
 var (
-	commit  = "unknown"
-	version = "unknown"
-	date    = "unknown"
-	dbapi   = "00824"
+	commit       = "unknown"
+	version      = "unknown"
+	date         = "unknown"
+	ConfigSchema = "00824"
+	SinkSchema   = "01110"
```
pashagolub:
```suggestion
	configSchema = "00824"
	sinkSchema   = "01110"
```

I think lower-cased var names are better to keep the whole style of the definition

internal/cmdopts/cmdconfig.go
```diff
@@ -104,8 +105,7 @@ func (cmd *ConfigUpgradeCommand) Execute([]string) (err error) {
 			}
 		}
 	} else {
-		opts.CompleteCommand(ExitCodeConfigError)
-		return errors.New("configuration storage does not support upgrade")
+		log.GetLogger(ctx).Warn("configuration storage does not support upgrade, skipping")
```
pashagolub: No, we shouldn't log here. We should return error. But let user ability to remove config option completely and re-run the command.

internal/cmdopts/cmdconfig.go
pashagolub: Because here we return error. We must behave the same in all situations.

The problem was not in how we handle, but that it's impossible to run upgrade command without specifying `--sources` and `--metrics` flags. This is the main issue! We must allow upgrade specifying only minimal `--sink` or `--metrics` options

Mohamed-Elfardy:
Thanks for the detailed feedback, that makes sense 
I understand now that the core issue is not just changing error → warning, but allowing upgrade to run with minimal flags
 (only --sink or --metrics) without requiring both --sources and --metrics.

I’ll rework the logic to make the behavior consistent across all cases and update the implementation accordingly.
Please let me know if there’s any preferred approach or specific place you’d like me to focus on first.

Mohamed-Elfardy:
All checks are passing now.

If there are any remaining adjustments you’d like me to make, please let me know and I’ll address them.

pashagolub:
overruled by #1175. Thanks for your contribution!

PR #1133, opened 2026-01-19
[+] add `Tables with the oldest non-frozen xid` panel

0xgouda:
- The panel displays the top $N tables with the oldest non-frozen transaction id, which helps identify tables that need vacuuming if transaction ids usage has gone high

Panel:

image
---

This is the first PR in a series of PRs that aim to improve the vacuum process panels.


0xgouda:
Also, I changed the Y axis in the `Transaction ID age` panel to use `unit: short` for easier readability:

image

Instead of:

image

pashagolub (approved):
LGTM!~

PR #1135, opened 2026-01-20
Add unit tests for types.go 

### LINKED ISSUE #1099: Increase test coverage
Our current test coverage is generally "low", observing the latest [coveralls builds](
https://coveralls.io/builds/77121244), we can see that the `reaper` and `sinks` packages are the lowest ones and hence should have more unit tests to increase their coverage.

The added tests should cover all scenarios; therefore, adding integration tests (which won't affect coverage) should also be considered.

abhijeetsharma200:
Fixes #1099
Part of multiple PRs to add unit and integration tests to improve coverage of `internal/sinks/`

### DIFF
--- a/grafana/postgres/v12/index-overview.json
+++ b/grafana/postgres/v12/index-overview.json
@@ -186,7 +186,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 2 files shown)


0xgouda:
Please re-read https://github.com/cybertec-postgresql/pgwatch/issues/1099#issuecomment-3769608366, our "low" test coverage clues that there are probably critical/important areas not covered by tests, hence we want to strengthen the tests for those, testing the `String()` method of the `SyncOp` is not one of these important areas, as it's already obvious that it's not buggy.

abhijeetsharma200:
Got it, thanks. I started with this as a quick warm-up and as a small guard against future refactors, but I agree it’s not a critical area. Will work on `internal/sinks/prometheus.go`.

PR #1137, opened 2026-01-21
[*] reorder `Health Check` panel

0xgouda:
- Reorder `Health Check` panel and group related panels under tabs.

image

image

### DIFF
--- a/grafana/postgres/v12/0-health-check.json
+++ b/grafana/postgres/v12/0-health-check.json
@@ -19,9 +19,22 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 21,
+  "id": 17,
   "links": [],
   "panels": [
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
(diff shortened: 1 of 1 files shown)


pashagolub:
Don't you think it breaks the whole idea of the Health Check dashboard? It supposed to be one huge "all-green" dashboard always displayable somewhere in the office for DBA team. Now every tab has 3 panels, it just doesn't make sense anymore. Plus in new Grafana those will be the [real tabs](https://grafana.com/mw/_next/image/?url=https%3A%2F%2Fs3.amazonaws.com%2Fa-us.storyblok.com%2Ff%2F1022730%2Fe99efe09a0%2Fdynamic-dashboards-navigation-grafana12.png&w=1152&q=75), not foldable groups like on your screenshot.

pashagolub (commented):
-

0xgouda:
> Don't you think it breaks the whole idea of the Health Check dashboard?

Ummm, thats a point. So I should remove the tabs (and revert panels to be 4 in each row) and retain the ordering?

pashagolub:
maybe we don't need to do nothing about it at all?

PR #1138, opened 2026-01-21
fix: sink-only schema upgrade now works (#1114)

### LINKED ISSUE #1114: Can't upgrade sink schema if using yaml source and metric configs

diiviikk5:
Fixes #1114

The upgrade command now accepts `--sink` alone when sources/metrics use yaml files.

Removed the validation that required both --sources and --metrics to be postgres URLs. Each config type is now upgraded independently if available.

### DIFF
--- a/internal/cmdopts/cmdconfig.go
+++ b/internal/cmdopts/cmdconfig.go
@@ -83,14 +83,26 @@ type ConfigUpgradeCommand struct {
 }
 
 // Execute upgrades the configuration schema.
+// Supports upgrading metrics/sources database, sink database, or both independently.
+// When using YAML files for sources/metrics, only the sink can be upgraded.
 func (cmd *ConfigUpgradeCommand) Execute([]string) (err error) {
 	opts := cmd.owner
-	if err = opts.ValidateConfig(); err != nil {
-		return
-	}
 	ctx := context.Background()
-	// Upgrade metrics/sources configuration if it's postgres
-	if opts.IsPgConnStr(opts.Metrics.Metrics) && opts.IsPgConnStr(opts.Sources.Sources) {
+
+	hasMetricsPg := opts.IsPgConnStr(opts.Metrics.Metrics)
+	hasSourcesPg := opts.IsPgConnStr(opts.Sources.Sources)
+	hasSinks := len(opts.Sinks.Sinks) > 0
+
+	// Require at least one upgradable target
+	if !hasMetricsPg && !hasSourcesPg && !hasSinks {
+		opts.CompleteCommand(ExitCodeConfigError)
+		return errors.New("no upgradable configuration specified: provide --sink or postgres connection strings for --sources/--metrics")
+	}
+
+	var upgraded bool
+
+	// Upgrade metrics if postgres
+	if hasMetricsPg {
 		err = opts.InitMetricReader(ctx)
 		if err != nil {
 			opts.CompleteCommand(ExitCodeConfigError)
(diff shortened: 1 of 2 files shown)


diiviikk5:
so afaik the config init command handles sources/metrics/sinks independently, but upgrade didn't ?. Is there a reason upgrade originally required both? Just curious

internal/cmdopts/cmdconfig.go
```diff
@@ -83,14 +83,26 @@ type ConfigUpgradeCommand struct {
...
+	// Require at least one upgradable target
+	if !hasMetricsPg && !hasSourcesPg && !hasSinks {
+		opts.CompleteCommand(ExitCodeConfigError)
+		return errors.New("no upgradable configuration specified: provide --sink or postgres connection strings for --sources/--metrics")
+	}
+
+	var upgraded bool
+
+	// Upgrade metrics/sources configuration if both are postgres
+	if hasMetricsPg && hasSourcesPg {
```
pashagolub: That's wrong. We must not depend on both options be the same of a kind. upgrade command must work with every option independently

pashagolub (changes requested):

internal/cmdopts/cmdconfig.go
diiviikk5: So  basically , - if metrics is postgres, upgrade it; if sources is postgres, upgrade it separately. ? Ill look into it

diiviikk5:
made them independent ,  @pashagolub , any other changes or feedback?

internal/cmdopts/cmdconfig.go
```diff
@@ -83,14 +83,26 @@ type ConfigUpgradeCommand struct {
...
-	if err = opts.ValidateConfig(); err != nil {
-		return
-	}
 	ctx := context.Background()
-	// Upgrade metrics/sources configuration if it's postgres
-	if opts.IsPgConnStr(opts.Metrics.Metrics) && opts.IsPgConnStr(opts.Sources.Sources) {
+
+	hasMetricsPg := opts.IsPgConnStr(opts.Metrics.Metrics)
+	hasSourcesPg := opts.IsPgConnStr(opts.Sources.Sources)
+	hasSinks := len(opts.Sinks.Sinks) > 0
```
pashagolub, line 94: Why only Postgres is here? What if later we want to upgrade YAML files from old format to a new one?

We should rely on checking if smth supports `metrics.Migrator` interface.

internal/cmdopts/cmdconfig.go
```diff
@@ -83,14 +83,26 @@ type ConfigUpgradeCommand struct {
 }
 
 // Execute upgrades the configuration schema.
+// Supports upgrading metrics/sources database, sink database, or both independently.
+// When using YAML files for sources/metrics, only the sink can be upgraded.
 func (cmd *ConfigUpgradeCommand) Execute([]string) (err error) {
 	opts := cmd.owner
-	if err = opts.ValidateConfig(); err != nil {
-		return
-	}
```
pashagolub: That's a correct idea! Validating is needed for a normal work but for a command we shall check in-place.

pashagolub:
also, please, do not overwrite my rebases ;-) 
> [diiviikk5](https://github.com/diiviikk5)(Divik Arora) [force-pushed](https://github.com/cybertec-postgresql/pgwatch/compare/25531be190a2e575135a5b25d04e9e94d7778d9e..51a709aa28e83b1295f271e1d7c4668e06cb7c41) the fix-sink-only-upgrade-1114 branch from [25531be](https://github.com/cybertec-postgresql/pgwatch/commit/25531be190a2e575135a5b25d04e9e94d7778d9e) to [51a709a](https://github.com/cybertec-postgresql/pgwatch/commit/51a709aa28e83b1295f271e1d7c4668e06cb7c41)

Please, make sure your branch is rebased against master

diiviikk5:
sorry couldnt update earlier  @pashagolub i was  ill ,but  thanks for reviewing the code properly it helps me learn ,  i did the rebase against the latest master  , Metrics, sources, and sinks are upgraded independently

pashagolub:
Test must pass

diiviikk5:
working on it , will inform / tag you once its done , will properly test it first before requesting workflow again

0xgouda:
overtaken by #1175

PR #1140, opened 2026-01-24
[-] fix data race in Prometheus sink, fixes #1136

### LINKED ISSUE #1136: Data race in PrometheusWriter(concurrent map iteration/write)
**Describe the bug**

I detected a race condition in the `PrometheusWriter` sink. The `Collect` method iterates over the global `promAsyncMetricCache` map without holding a lock, while metric updates may concurrently arrive via `Write → PromAsyncCacheAddMetricData`. Since the cache is a Go map, this leads to unsynchronized concurrent iteration and mutation.

**To Reproduce**
Run the following test with the race detector enabled:

```
package sinks

import (
	"context"
	"sync"
	"testing"
	"time"

	"github.com/cybertec-postgresql/pgwatch/v5/internal/metrics"
	"github.com/prometheus/client_golang/prometheus"
)

func TestCollect_RaceCondition(t *testing.T) {
	// 1. Initialize the real PrometheusWriter
	// Note: In the current buggy code, this shares the global 'promAsyncMetricCache'
	promw, _ := NewPrometheusWriter(context.Background(), "127.0.0.1:0/pgwatch")

	// 2. Register a metric so Write() actually puts data into the map
	promw.SyncMetric("race_db", "test_metric", AddOp)

	var wg sync.WaitGroup
	done := make(chan struct{})

	// --- The Writer (Simulating Database Updates) ---
	wg.Add(1)
	go func() {
		defer wg.Done()
		for {
			select {
			case <-done:
				return
			default:
				// Call the Write method
				promw.Write(metrics.MeasurementEnvelope{
					DBName:     "race_db",
					MetricName: "test_metric",
					Data: metrics.Measurements{
						{
							metrics.EpochColumnName: time.Now().UnixNano(),
							"value":                 int64(100),
						},
					},
				})
			}
		}
	}()

	// --- The Collector (Simulating Prometheus Scrapes) ---
	wg.Add(1)
	go func() {
		defer wg.Done()
		ch := make(chan prometheus.Metric, 10000)

		for i := 0; i < 50; i++ {
			promw.Collect(ch)

			// Drain the channel so it doesn't block
		drainLoop:
			for {
				select {
				case <-ch:
				default:
					break drainLoop
				}
			}
		}
		close(done)
	}()

	wg.Wait()
}
```

**Expected behavior**

Concurrent calls to `Write()` and `Collect()` should be thread-safe and not have a data race.

**Actual behavior**

```
WARNING: DATA RACE
Read at ... by goroutine 31:
  runtime.mapIterStart()
  .../internal/sinks/prometheus.go:171

Previous write at ... by goroutine 30:
  runtime.mapaccess2_faststr()
  .../internal/sinks/prometheus.go:119
testing.go:1617: race detected during execution of test


```

abhijeetsharma200:
Fixes #1136. 

Will add the test as part of the PR to add tests to prometheus.go

### DIFF
--- a/internal/sinks/prometheus_race_test.go
+++ b/internal/sinks/prometheus_race_test.go
@@ -1,19 +1,19 @@
 package sinks
 
 import (
-	"context"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/cybertec-postgresql/pgwatch/v5/internal/metrics"
+	"github.com/cybertec-postgresql/pgwatch/v5/internal/testutil"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
 func TestCollect_RaceCondition_Real(_ *testing.T) {
 	// 1. Initialize the real PrometheusWriter
 	// Note: In the current buggy code, this shares the global 'promAsyncMetricCache'
-	promw, _ := NewPrometheusWriter(context.Background(), "127.0.0.1:0/pgwatch")
+	promw, _ := NewPrometheusWriter(testutil.TestContext, "127.0.0.1:0/pgwatch")
 
 	// 2. Register a metric so Write() actually puts data into the map
 	_ = promw.SyncMetric("race_db", "test_metric", AddOp)
@@ -22,9 +22,7 @@ func TestCollect_RaceCondition_Real(_ *testing.T) {
 	done := make(chan struct{})
 
 	// --- The Writer (Simulating Database Updates) ---
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
+	wg.Go(func() {
 		for {
 			select {
 			case <-done:
@@ -44,17 +42,15 @@ func TestCollect_RaceCondition_Real(_ *testing.T) {
 				// No sleep here -> hammer the map as fast as possible
 			}
 		}
-	}()
+	})
 
 	// --- The Collector (Simulating Prometheus Scrapes) ---
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
+	wg.Go(func() {
 		// Prometheus provides a channel to receive metrics
 		ch := make(chan prometheus.Metric, 10000)
 
 		// Scrape 50 times (more than enough to trigger a race in a tight loop)
-		for i := 0; i < 50; i++ {
+		for range 50 {
 			// Call the REAL Collect method
 			promw.Collect(ch)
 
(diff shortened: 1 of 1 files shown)


internal/sinks/prometheus.go
```diff
@@ -159,15 +159,28 @@ func (promw *PrometheusWriter) Collect(ch chan<- prometheus.Metric) {
...
 	if len(promAsyncMetricCache) == 0 {
+		promAsyncMetricCacheLock.Unlock()
 		promw.logger.Warning("No dbs configured for monitoring. Check config")
 		ch <- promw.totalScrapeFailures
 		promw.lastScrapeErrors.Set(0)
 		ch <- promw.lastScrapeErrors
 		return
 	}
+
+	snapshot := promAsyncMetricCache
```
pashagolub, line 172: I'd prefer to have a longer lock rather than duplicating metric cache that might be huge on high loaded systems

abhijeetsharma200: It's not a deep copy; just a pointer swap (O(1), which is 8/4 bytes depending on OS). The map data stays in place, we just change which pointer the global variable holds. After processing, the old map gets garbage collected. New writes go to the fresh map and get picked up on the next scrape. There shouldn't be any data loss because prometheus is scraping periodically

abhijeetsharma200: Also, I noticed `promAsyncMetricCache` is a package-level global rather than a field on the PrometheusWriter struct. Is there a specific reason for that design? Just curious if multiple writer instances are expected to share the cache intentionally

pashagolub: old design. should be a part of a struct actually

abhijeetsharma200:
@pashagolub ping for review

pashagolub:
Could you please add a test case that must fail before the change and is working now. Thanks

abhijeetsharma200:
> Could you please add a test case that must fail before the change and is working now. Thanks

Done! Please let me know if there any additional changes necessary

internal/sinks/prometheus.go
abhijeetsharma200: > old design. should be a part of a struct actually

I see, can I then open an issue to refactor it and make it part of a struct?

pashagolub:
Thanks!

abhijeetsharma200:
> old design. should be a part of a struct actually

I see, can I then open an issue to refactor it and make it part of a struct?

pashagolub:
absolutely. be my guest! :-)

PR #1146, opened 2026-01-26
[+] improve `Index overview` dashboard

0xgouda:
1. Add panels that utilize the columns that, for some reason, were not used in the `index_stats` metric:
    - `Top n indexes by disk reads (blocks)`
    - `Top n indexes by cache hits (blocks)`
    - `Top n indexes by tuples read`
    - `Top n indexes by tuples fetched`
2. Update the `Top n biggest indexes` to return the number of scans that happened in the specified time interval instead of the total scans since index creation.

Screenshots:
image

image

image

### DIFF
--- a/grafana/postgres/v12/index-overview.json
+++ b/grafana/postgres/v12/index-overview.json
@@ -596,6 +596,7 @@
           }
         }
       ],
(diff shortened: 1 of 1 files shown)


pashagolub (approved):
Great!

PR #1154, opened 2026-01-31
[+] more improvements to `index overview` dashboard 

0xgouda:
- Update the newly added `Top $n indexes by reads/hits` in terms of bytes per second by joining the `block_size` from the settings metric
- Refactor the `reco_add_index` to return the query IDs affected by the index creation in case users want to investigate them further
- Refactor `reco_add_index` to return the top queries by `sum(avg_exec_time)` for all affected queries, not by the heaviest query improved.

Screenshots:

image

image

### DIFF
--- a/internal/metrics/metrics.yaml
+++ b/internal/metrics/metrics.yaml
@@ -1400,6 +1400,16 @@ metrics:
                         json_array_elements_text(indexes->'queryids') as query_id
                     where
                         indexes->>'ddl' > ''
+                ), aggregated_stat_statements as (
+                    select 
+                        queryid,
+                        sum(total_exec_time) as total_exec_time,
+                        sum(calls) as calls
+                    from 
+                        pg_stat_statements
+                    where 
+                        queryid in (select query_id from qrs)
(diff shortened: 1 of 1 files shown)


internal/metrics/metrics.yaml
```diff
@@ -1391,28 +1391,30 @@ metrics:
...
+                select
                     (extract(epoch from now()) * 1e9)::int8 as epoch_ns,
                     object_name as tag_object_name,
                     'create_index'::text as tag_reco_topic,
-                    qrs.query_id,
-                    qrs.affected_query_count,
+                    max(qrs.affected_query_ids) as affected_query_ids,
                     qrs.recommendation,
-                    s.total_exec_time
+                    sum(s.total_exec_time / s.calls) as total_avg_exec_time
```
0xgouda, line 1420: After rethinking it, I saw that it's better to consider the sum of the average runtime for all executed queries to order the most important suggestions instead of the total runtime of the heaviest query.

What do you think @pashagolub?

PR #1155, opened 2026-01-31
feat: consistent sorting for metrics presets

### LINKED ISSUE #1050: Sort presets and show description
When adding a source, the current UI isn't very helpful:

Image

"Basic" is preselected, that is good, but then there's "exhaustive", "minimal" and several cloud options in random (?) order. It would be nice if the options had some more logical ordering like
* minimal
* basic
* ...
* exhaustive
* cloudthing1
* cloudthing2
* cloudthing2_extensive
* special developer thing
* ...

In the presets menu, the presets have descriptions, perhaps that could be shown there as well?

sahaj-b:
Fixes #1050, Adds sort_order field to presets to ensure consistent ordering in the WebUI.
### Changes Made
- **YAML**: Added `sort_order` field to all 16 built-in presets in `metrics.yaml`
- **Backend**: 
  - Updated`Preset` struct, and updated SQL queries to include `sort_order`
  - Created migration to add `sort_order` column to `presets` table with hardcoded sort_order values
- **Frontend**: 
  - Updated preset dropdown and grid to use sort_order for sorting
  -  Added description popover on hover for Presets and metrics dropdown
  
Order:
```
1. minimal
2. basic
3. standard
4. exhaustive
5. full
6. aiven
7. azure
8. gce
9. rds
10. pgbouncer
11. pgpool
12. unprivileged
13. recommendations
14. prometheus-async
15. exhaustive_no_python
16. debug
```

### DIFF
--- a/internal/webui/src/pages/PresetsPage/components/PresetsGrid/PresetsGrid.tsx
+++ b/internal/webui/src/pages/PresetsPage/components/PresetsGrid/PresetsGrid.tsx
@@ -1,7 +1,12 @@
-import { useMemo, useState, useEffect, useCallback } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
+import ArrowDownwardIcon from "@mui/icons-material/ArrowDownward";
+import ArrowUpwardIcon from "@mui/icons-material/ArrowUpward";
+import { Box, IconButton } from "@mui/material";
 import { DataGrid, GridColDef } from "@mui/x-data-grid";
+import { useQueryClient } from "@tanstack/react-query";
 import { Error as ErrorComponent } from "components/Error/Error";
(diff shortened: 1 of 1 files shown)


sahaj-b:
I can also simply hardcode the order in the API route handler in `internal/webserver/metric.go` instead of adding a sort_order field which requires a migration. 
But I felt having it in YAML/DB is more flexible for future changes.
Let me know if you prefer hardcoding it.

sahaj-b:
Also, the migration doesn't work due to this issue #1156

sahaj-b:
Popover description
image

Order column
image

pashagolub:
No, I don't want this. For such a small issue to produce such a huge amount of code.

This must be solved purely (sic!) on frontend side

PR #1157, opened 2026-02-01
fix: config upgrade circular dependency

### LINKED ISSUE #1156: Bug: `pgwatch config upgrade` cannot run when schema is outdated, circular dependency.
**Description**
The pgwatch config upgrade command fails to execute migrations because it cannot initialize the database connection when migrations are pending. The command tells users to run itself, creating a circular dependency that makes migrations impossible to execute via CLI.

**Steps to Reproduce**
1. Start with old schema:
`docker compose -f ./docker/docker-compose.yml up -d`.

2. Stop pgwatch (keep postgres running with old schema):
`docker compose -f ./docker/docker-compose.yml stop pgwatch`.

3. Add new migration by editing `internal/metrics/postgres_schema.sql` and `internal/metrics/postgres_schema.go` 

3. Build new image:
`docker compose -f ./docker/docker-compose.yml build pgwatch`.

4. Try to run migration:
```sh
docker run --rm --network pgwatch_default \
  -v $(pwd)/internal/metrics/metrics.yaml:/etc/pgwatch/metrics.yml \
  -v ./docker/bootstrap/.pgpass:/root/.pgpass:ro \
  cybertecpostgresql/pgwatch:latest \
  config upgrade \
  --sources=postgresql://pgwatch@postgres:5432/pgwatch \
  --metrics=postgresql://pgwatch@postgres:5432/pgwatch
```

**Expected**: Migration runs and updates schema
**Actual**: ```config database schema is outdated, please run migrations using `pgwatch config upgrade` command```

**Root Cause**
The `config upgrade` command (in `internal/cmdopts/cmdconfig.go`) has circular logic:
1. **Line 94:** Calls `opts.InitMetricReader(ctx)` to initialize the database connection
2. `InitMetricReader()` -> `NewPostgresMetricReaderWriterConn()` (in `internal/metrics/postgres.go`)
3. **Line 29-32:** `NewPostgresMetricReaderWriterConn()` checks `NeedsMigration()` and returns `ErrNeedsMigration` if pending migrations exist
4. **Line 96-97:** catches the error and exits
5. **Line 100** (`m.Migrate()`) is **never reached**

sahaj-b:
Fixes #1156 

Created dedicated migration-only constructors that skip the migration check:
1. Added `NewPostgresMetricMigrator()` and `NewPostgresSinkMigrator()`
2. Updated `Execute()` in `internal/cmdopts/cmdconfig.go` to use the new migrator constructors instead of full reader/writer initialization

### DIFF
--- a/internal/cmdopts/cmdconfig.go
+++ b/internal/cmdopts/cmdconfig.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 
 	"github.com/cybertec-postgresql/pgwatch/v5/internal/metrics"
+	"github.com/cybertec-postgresql/pgwatch/v5/internal/sinks"
 	"github.com/cybertec-postgresql/pgwatch/v5/internal/sources"
 )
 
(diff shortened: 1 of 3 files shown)


internal/cmdopts/cmdconfig.go
```diff
@@ -91,38 +92,31 @@ func (cmd *ConfigUpgradeCommand) Execute([]string) (err error) {
...
+			return
 		}
 	} else {
 		opts.CompleteCommand(ExitCodeConfigError)
 		return errors.New("configuration storage does not support upgrade")
 	}
 	// Upgrade sinks configuration if it's postgres
 	if len(opts.Sinks.Sinks) > 0 {
-		err = opts.InitSinkWriter(ctx)
+		var m metrics.Migrator
```
pashagolub, line 111: what if it's not a postgres sink?

pashagolub:
thanks for your contribution! Bug managed in #1171

PR #1160, opened 2026-02-02
[*] remove redundant `Source Name` from `Tables Overview` dashboard, fixes #1134

### LINKED ISSUE #1134: Remove Redundant `Source Name` field
There is no point in having the dbname/"Source Name" field repeated in dashboards that only accept a single value for the dbname/"Source Name". Also, from a performance perspective, removing it will decrease the amount of network I/O a little bit
Image

Mohamed-Elfardy:
Fixes :  #1134
This PR removes the redundant dbname / "Source Name" field from dashboards
that only accept a single dbname value to be more clarify.

The change is limited to single-db dashboards (e.g. tables-overview.json).

Global dashboards that may display multiple databases are intentionally left unchanged.
If there are other single-db dashboards should be updated, please let me know.

### DIFF
--- a/grafana/postgres/v12/4-tables-overview.json
+++ b/grafana/postgres/v12/4-tables-overview.json
@@ -1457,7 +1457,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
-          "rawSql": "SELECT\n*\nFROM\n(\n  SELECT DISTINCT ON (table_full_name)\n  tag_data->>'table_full_name' as table_full_name,\n  data->>'tx_freeze_age' as tx_freeze_age,\n  (data->>'seconds_since_last_vacuum')::int8  as last_vacuum,\n  (data->>'seconds_since_last_analyze')::int8 as last_analyze\n  FROM\n    table_stats\n  WHERE\n    $__timeFilter(time)\n    AND dbname = '$dbname'\n  ORDER BY\n    1, time DESC\n)\nORDER BY 3 DESC NULLS LAST\nLIMIT $top\n",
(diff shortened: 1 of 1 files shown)


0xgouda:
It's not removed; you just removed the aliasing. You'd also need not to query it in the panels.

image

Mohamed-Elfardy:
Just to confirm my understanding : for single-db dashboards, should the dbname column be removed entirely from the panels, relying on the dashboard context instead , right ?

grafana/postgres/v12/4-tables-overview.json
```diff
@@ -200,7 +188,7 @@
...
+          "rawSql": "select\n  x.*,\n  (data->>'seconds_since_last_vacuum')::int8 as last_vacuum,\n  (data->>'seconds_since_last_analyze')::int8 as last_analyze\nfrom (\nSELECT\n  tag_data->>'table_full_name' as table_full_name,\n  max((data->>'total_relation_size_b')::int8) as total_relation_size,\n  max((data->>'table_size_b')::int8) as data_size\nFROM\n  table_stats\nWHERE\n  $__timeFilter(time)\n  AND dbname = '$dbname'\nGROUP BY 1 \nORDER BY 3 DESC NULLS LAST\nLIMIT $top\n) x\njoin lateral (select data from table_stats where dbname = '$dbname' and tag_data->>'table_full_name' = x.table_full_name and $__timeFilter(time) order by time desc limit 1) y on true",
```
0xgouda: Why haven't you updated the `ORDER BY` here?

Please ensure that `ORDER BY` clauses are updated accordingly in all panels.

0xgouda:
Plus: resize the panels after removing the `Source Name` column and make them align left.

also add insepct option to Table name column.

Mohamed-Elfardy:
> and make them align left.

Screenshot 2026-02-04 014846

I changed the default alignment to left for the panel.
Is this acceptable, or would you prefer keeping the default as auto and only setting left alignment via column overrides
( like in this screen )?

Mohamed-Elfardy:
> also add insepct option to Table name column.

I enabled inspect via a field override on the Table Name column. Should this be done via overrides, or by setting the default inspect option to true?

0xgouda:
> I enabled inspect via a field override on the Table Name column. Should this be done via overrides, or by setting the default inspect option to true?

It should be an override since we want it for the `Table Name` field only, good job!

Mohamed-Elfardy:
> It should be an override since we want it for the `Table Name` field only, good job!

got it , is there any suggested additions, please let me know

PR #1164, opened 2026-02-02
feat(cli): add structured output formatting and file redirection

Amr-Shams:
- Added --output flag supporting 'json' and 'text' (default) formats.
    - Added --list flag for both metric and source to list metrics/sources
    - Added unit test for both subcommands
    - Refactored all the current code output to be redirected to the formater


pashagolub:
Managed in #1165

Amr-Shams:
> Managed in #1165

few mins late 🤣 , what about the --output for formatting ?

pashagolub:
Well, I do not see any use case for `--output`

PR #1175, opened 2026-02-03
[+] allow sink schema upgrades without --metrics/--sources, fixes #1114

### LINKED ISSUE #1114: Can't upgrade sink schema if using yaml source and metric configs

pashagolub:
Removed the requirement that both sources and metrics must be PostgreSQL connections. Added support for upgrading only sink when sources/metrics are YAML files. Changed error messages to warnings ([WARN]) for non-upgradeable configurations. Only fails if no components support upgrade (previously failed if any didn't support upgrade).

### DIFF
--- a/internal/cmdopts/cmdconfig.go
+++ b/internal/cmdopts/cmdconfig.go
@@ -95,59 +95,55 @@ func (cmd *ConfigUpgradeCommand) Execute([]string) (err error) {
 	}
 
 	ctx := context.Background()
-	var upgraded bool // track if any component was upgraded
 
-	// Upgrade metrics/sources configuration if it's postgres
-	if opts.IsPgConnStr(opts.Metrics.Metrics) && opts.IsPgConnStr(opts.Sources.Sources) {
-		if opts.MetricsReaderWriter == nil {
-			err = opts.InitMetricReader(ctx)
-			if err != nil {
-				opts.CompleteCommand(ExitCodeConfigError)
-				return
-			}
+	f := func(uri string, newMigratorFunc func() (any, error)) error {
+		if uri == "" {
+			return nil
 		}
-		if m, ok := opts.MetricsReaderWriter.(db.Migrator); ok {
-			err = m.Migrate()
-			if err != nil {
-				opts.CompleteCommand(ExitCodeConfigError)
-				return
-			}
-			upgraded = true
-		} else {
-			fmt.Fprintln(opts.OutputWriter, "[WARN] configuration storage does not support upgrade, skipping")
+		if !opts.IsPgConnStr(uri) {
+			return fmt.Errorf("cannot upgrade storage %s: %w", uri, errors.ErrUnsupported)
 		}
-	} else if opts.Metrics.Metrics > "" || opts.Sources.Sources > "" {
-		// At least one is specified but not postgres connection - log warning
-		fmt.Fprintln(opts.OutputWriter, "[WARN] configuration storage does not support upgrade, skipping")
+		m, initErr := newMigratorFunc()
+		if initErr != nil {
+			return initErr
+		}
+		return m.(db.Migrator).Migrate()
+
 	}
 
-	// Upgrade sinks configuration if it's postgres
-	if len(opts.Sinks.Sinks) > 0 {
-		if opts.SinksWriter == nil {
-			opts.SinksWriter, err = sinks.NewSinkWriter(ctx, &opts.Sinks)
-			if err != nil {
-				opts.CompleteCommand(ExitCodeConfigError)
-				return
-			}
-		}
-		if m, ok := opts.SinksWriter.(db.Migrator); ok {
-			err = m.Migrate()
-			if err != nil {
-				opts.CompleteCommand(ExitCodeConfigError)
-				return
-			}
-			upgraded = true
-		} else {
-			fmt.Fprintln(opts.OutputWriter, "[WARN] sink storage does not support upgrade, skipping")
-		}
+	err = f(opts.Sources.Sources, func() (any, error) {
+		return sources.NewPostgresSourcesReaderWriter(ctx, opts.Sources.Sources)
+	})
+
+	err = errors.Join(err, f(opts.Metrics.Metrics, func() (any, error) {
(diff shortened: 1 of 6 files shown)


internal/cmdopts/cmdconfig.go
```diff
@@ -87,46 +88,66 @@ type ConfigUpgradeCommand struct {
...
+		} else {
+			fmt.Fprintln(opts.OutputWriter, "[WARN] configuration storage does not support upgrade, skipping")
 		}
-	} else {
-		opts.CompleteCommand(ExitCodeConfigError)
-		return errors.New("configuration storage does not support upgrade")
+	} else if opts.Metrics.Metrics > "" || opts.Sources.Sources > "" {
+		// At least one is specified but not postgres connection - log warning
+		fmt.Fprintln(opts.OutputWriter, "[WARN] configuration storage does not support upgrade, skipping")
 	}
```
0xgouda: Why do we require `opts.IsPgConnStr(opts.Metrics.Metrics) && opts.IsPgConnStr(opts.Sources.Sources)` to upgrade metrics config?

What if we are using pg conn for metrics but yaml for sources? Why not upgrade then?

Actually, I am not sure why sources are considered at all, since we are only using `metricsReaderWriter`. Is it for the case where `--metrics` is not specified, so it defaults to using the pg conn source?

Also, why don't we run migrations for `sourcesReaderWriter`? its not used at all.

0xgouda:
Also, when upgrading from `v4.1` to `v5`, now the CLI states that we should run `config upgrade`, which succeeds, but without changing the `admin` schema. Maybe we should add a new sink migration for changing the functions that got updates in the admin schema?

```
$ go run ./cmd/pgwatch --sources="contrib/sample.sources.yaml" --sink="postgres://postgres:postgres@localhost:5437/postgres" --metrics="internal/metrics/metrics.yaml" 
2026-02-04 02:43:30.882 [INFO] [sink:postgres] [db:postgres] initialising measurements database...
2026-02-04 02:43:30.891 [INFO] [sink:postgres] [db:postgres] measurements sink is activated
2026-02-04 02:43:30.891 [ERROR] sink database schema is outdated, please run migrations using `pgwatch config upgrade` command
exit status 1

$ go run ./cmd/pgwatch --sources="contrib/sample.sources.yaml" --sink="postgres://postgres:postgres@localhost:5437/postgres" --metrics="internal/metrics/metrics.yaml" config upgrade
[WARN] configuration storage does not support upgrade, skipping
2026-02-04 02:43:35.098 [INFO] [sink:postgres] [db:postgres] initialising measurements database...
2026-02-04 02:43:35.107 [INFO] [sink:postgres] [db:postgres] measurements sink is activated
2026-02-04 02:43:35.117 [INFO] [sink:postgres] [db:postgres] Applying migration named '01110 Apply postgres sink schema migrations'...
2026-02-04 02:43:35.118 [INFO] [sink:postgres] [db:postgres] Applied migration named '01110 Apply postgres sink schema migrations'

$ go run ./cmd/pgwatch --sources="contrib/sample.sources.yaml" --sink="postgres://postgres:postgres@localhost:5437/postgres" --metrics="internal/metrics/metrics.yaml"               
2026-02-04 02:43:39.332 [INFO] [sink:postgres] [db:postgres] initialising measurements database...
2026-02-04 02:43:39.341 [INFO] [sink:postgres] [db:postgres] measurements sink is activated
2026-02-04 02:43:39.342 [INFO] [sources:1] sources refreshed
2026-02-04 02:43:39.346 [INFO] [metrics:74] [presets:14] metrics and presets refreshed
2026-02-04 02:43:39.352 [INFO] [source:omdb.test] [recovery:false] Connect OK. Version: PostgreSQL 18.1 (Ubuntu 18.1-1.pgdg24.04+2) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, 64-bit
2026-02-04 02:43:39.352 [INFO] [source:omdb.test] [metric:wal] [interval:60] starting gatherer
2026-02-04 02:43:39.353 [INFO] [source:omdb.test] [metric:wal] [rows:1] [cache:false] measurements fetched
2026-02-04 02:43:40.296 [ERROR] [sink:postgres] [db:postgres] [sql:select * from admin.ensure_partition_metric_dbname_time($1, $2, $3, $4)] [args:[wal omdb.test 2026-02-04 02:43:39.353655 +0200 EET 1 week]] [err:ERROR: cross-database references are not implemented: "subpartitions.wal_omdb.test" (SQLSTATE 0A000)] [pid:46954] [time:4.732447ms] Query
```

internal/cmdopts/cmdconfig.go
pashagolub: yeah.... this is legacy :) So, back in times we store both metrics ans sources in db. And this is was just config. Then we decided to allow to use custom metrics from file. Now sources are in db, but metrics might be in YAML. Then we add support for both sources and metrics in YAML. 

But the thing is if one combines DB and YAML, database will have the full schema including metrics and sources. So if we update even one, we automatically update both. That's why we don't have a separate migrations for `sourcesReaderWriter`, it's automatically covered by `metricsReaderWriter`

0xgouda: so it should be `||` not `&&` right? because currently it won't upgrade if both are not pg sources

Also, I am thinking if a user for some reason decided to use 2 dbs, one for metrics and the other for sources, this would be a bug, right? because `metricsReaderWriter` will only update the db used for metrics not ther other

pashagolub: hmmm, yeah, it's possible. and users do stupid things. :) Thanks, I'll handle every one of them independently

pashagolub:
> Also, when upgrading from `v4.1` to `v5`, now the CLI states that we should run `config upgrade`, which succeeds, but without changing the `admin` schema. Maybe we should add a new sink migration for changing the functions that got updates in the admin schema?
> 

I thought they were added already... Would you please add those in a separate PR to mention the fact of migration. thanks

PR #1181, opened 2026-02-04
[+] apply admin functions migrations for v5, fixes #1180

### LINKED ISSUE #1180: Apply admin functions migrations for v5

0xgouda:
closes #1180

Tested sink migrations from v3.4.0 and upwards to v5, and all worked correctly

### DIFF
--- a/cmd/pgwatch/version.go
+++ b/cmd/pgwatch/version.go
@@ -8,7 +8,7 @@ var (
 	version      = "unknown"
 	date         = "unknown"
 	configSchema = "00824"
-	sinkSchema   = "01110"
+	sinkSchema   = "01180"
 )
 
 func printVersion() {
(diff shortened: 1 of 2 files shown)


internal/sinks/postgres.go
```diff
@@ -603,6 +603,23 @@ var migrations func() migrator.Option = func() migrator.Option {
...
+			Func: func(ctx context.Context, tx pgx.Tx) error {
+				_, err := tx.Exec(ctx, `DROP FUNCTION IF EXISTS admin.ensure_partition_metric_dbname_time`)
+				if err != nil {
+					return err
+				}
+
+				_, err = tx.Exec(ctx, sqlMetricEnsurePartitionPostgres)
+				if err != nil {
+					return err
+				}
```
0xgouda, line 617: Added `DROP FUNCTION IF EXISTS admin.ensure_partition_metric_dbname_time` for users migrating from `v4.0.0` to v5, as in `v4.1` we forgot to add the migration for that.

And it will be re-created with the v5 migration in script `sqlMetricEnsurePartitionPostgres`

PR #1184, opened 2026-02-06
[#1092] Enhance metric form UI

### LINKED ISSUE #1092: Improvement for new metric form in UI

Mazen050:
Closes #1092 

This PR Enhances the look of the metric form UI by introducing 2 individual boxs one for postgres version number and the other for sql, with another button to add New SQL.

Examples of how it looks:

Screenshot 2026-02-06 235227

image

Also versions dropdown is editable and can be used to add a new versions, and it also successfully retrieves older metric sql from backend.

### DIFF
--- a/internal/webui/src/containers/MetricFormDialog/MetricFormDialog.consts.ts
+++ b/internal/webui/src/containers/MetricFormDialog/MetricFormDialog.consts.ts
@@ -14,21 +14,16 @@ export const getMetricInitialValues = (data?: MetricGridRow): MetricFormValues =
     Gauges: convertGauges(data?.Metric.Gauges ?? [""]),
     InitSQL: data?.Metric.InitSQL ?? "",
     IsInstanceLevel: data?.Metric.IsInstanceLevel ?? false,
-    SQLs: yaml.stringify(data?.Metric.SQLs) ?? "",
+    SQLs: (data?.Metric.SQLs) ?? "",
   };
 };
 
 export const createMetricRequest = (values: MetricFormValues): MetricRequestBody => {
-  const sqls: Record<number, string> = {};
-  yaml.parse(values.SQLs, (key, value) => {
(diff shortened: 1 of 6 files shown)


0xgouda:
Thanks, will take a look tomorrow.

Mazen050:
Please note that I have added a `TODO` in the form to restrict input to numbers and dots maybe in the future or maybe in this PR depends if it's needed or not.

Mazen050:
I have made the version input allow only numbers and dots.

pashagolub:
thanks for working on this! I have doubts if such UI is reasonably good for the current metrics that are usually several screens long. From the UX point of view it's impossible to make any sane actions with such a layout and lack of syntax highlightning

Mazen050:
Would making the text area bigger be any better? Currently it starts with 3 rows and expands to 12 rows and then it changes to scrolling. 

So we can change the number of max rows or we can make the sql text area wider too since the version input doesn't need that much space.

About syntax highlighting we could use something like a codemirror component instead of the text area.

Mazen050:
Hello, Can I know the status here please? If the feature is still needed I would gladly continue working on this.

Although if the feature is not needed anymore we can close this PR.

Mazen050:
@pashagolub is the feature not needed or is something wrong with the PR?

pashagolub:
Nothing wrong with PR. The UI should look different. I don't know how yet.

PR #1186, opened 2026-02-08
[*] refactor `TryCreateMissingExtensions` 

gemy26:
- Refactored `TryCreateMissingExtensions` to use direct pgx connection instead of the `QueryMeasurements` wrapper.

### DIFF
--- a/internal/reaper/database.go
+++ b/internal/reaper/database.go
@@ -473,35 +473,54 @@ func (r *Reaper) GetObjectChangesMeasurement(ctx context.Context, md *sources.So
 // whitelisting or "trusted extensions" (a feature from v13). Ignores errors.
 func TryCreateMissingExtensions(ctx context.Context, md *sources.SourceConn, extensionNames []string, existingExtensions map[string]int) []string {
 	// TODO: move to sources package and use direct pgx connection
-	sqlAvailable := `select name::text from pg_available_extensions`
+	var queryArgs []any
+	l := log.GetLogger(ctx)
 	extsCreated := make([]string, 0)
 
-	// For security reasons don't allow to execute random strings but check that it's an existing extension
-	data, err := QueryMeasurements(ctx, md, sqlAvailable)
-	if err != nil {
-		log.GetLogger(ctx).Infof("[%s] Failed to get a list of available extensions: %v", md, err)
-		return extsCreated
+	if !md.IsPostgresSource() {
+		queryArgs = append(queryArgs, pgx.QueryExecModeSimpleProtocol)
 	}
 
-	availableExts := make(map[string]bool)
-	for _, row := range data {
-		availableExts[row["name"].(string)] = true
+	sqlAvailable := `select name::text from pg_available_extensions`
+
+	// Use direct pgx Query instead of QueryMeasurements wrapper
+	rows, err := md.Conn.Query(ctx, sqlAvailable, queryArgs...)
+	if err != nil {
+		l.Infof("[%s] Failed to get a list of available extensions: %v", md, err)
+		return extsCreated
+	}
+	defer rows.Close()
+
+	// For security reasons don't allow to execute random strings but check that it's an existing extension
+	availableExtentions := make(map[string]bool)
+	var extentionName string
+	for rows.Next() {
+		if err := rows.Scan(&extentionName); err != nil {
+			l.Infof("[%s] Failed to scan extension name: %v", md, err)
+			continue
+		}
+		availableExtentions[extentionName] = true
 	}
 
 	for _, extToCreate := range extensionNames {
 		if _, ok := existingExtensions[extToCreate]; ok {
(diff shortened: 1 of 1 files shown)


internal/reaper/database.go
```diff
@@ -473,35 +473,54 @@ func (r *Reaper) GetObjectChangesMeasurement(ctx context.Context, md *sources.So
...
 func TryCreateMissingExtensions(ctx context.Context, md *sources.SourceConn, extensionNames []string, existingExtensions map[string]int) []string {
 	// TODO: move to sources package and use direct pgx connection
-	sqlAvailable := `select name::text from pg_available_extensions`
+	var queryArgs []any
+	l := log.GetLogger(ctx)
 	extsCreated := make([]string, 0)
 
-	// For security reasons don't allow to execute random strings but check that it's an existing extension
-	data, err := QueryMeasurements(ctx, md, sqlAvailable)
+	if !md.IsPostgresSource() {
```
pashagolub, line 480: if it's not a postgres source, there is no any extensions. we can return immediately

internal/reaper/database.go
```diff
@@ -473,35 +473,54 @@ func (r *Reaper) GetObjectChangesMeasurement(ctx context.Context, md *sources.So
...
+		queryArgs = append(queryArgs, pgx.QueryExecModeSimpleProtocol)
+	}
+
+	sqlAvailable := `select name::text from pg_available_extensions`
+
+	// Use direct pgx Query instead of QueryMeasurements wrapper
+	rows, err := md.Conn.Query(ctx, sqlAvailable, queryArgs...)
 	if err != nil {
-		log.GetLogger(ctx).Infof("[%s] Failed to get a list of available extensions: %v", md, err)
+		l.Infof("[%s] Failed to get a list of available extensions: %v", md, err)
```
pashagolub, line 489: that's not how we form logging messages

internal/reaper/database.go
```diff
@@ -473,35 +473,54 @@ func (r *Reaper) GetObjectChangesMeasurement(ctx context.Context, md *sources.So
...
-		availableExts[row["name"].(string)] = true
+	// For security reasons don't allow to execute random strings but check that it's an existing extension
+	availableExtentions := make(map[string]bool)
+	var extentionName string
+	for rows.Next() {
+		if err := rows.Scan(&extentionName); err != nil {
+			l.Infof("[%s] Failed to scan extension name: %v", md, err)
+			continue
+		}
+		availableExtentions[extentionName] = true
```
pashagolub, line 502: I think it's easier to use pgx.CollectRows functionality https://pkg.go.dev/github.com/jackc/pgx/v5#CollectRows

internal/reaper/database.go
```diff
@@ -473,35 +473,54 @@ func (r *Reaper) GetObjectChangesMeasurement(ctx context.Context, md *sources.So
...
+		l.Infof("[%s] Failed to get a list of available extensions: %v", md, err)
 		return extsCreated
 	}
+	defer rows.Close()
 
-	availableExts := make(map[string]bool)
-	for _, row := range data {
-		availableExts[row["name"].(string)] = true
+	// For security reasons don't allow to execute random strings but check that it's an existing extension
+	availableExtentions := make(map[string]bool)
```
pashagolub, line 495: Why do we need `map[string]bool` here?

internal/reaper/database.go
```diff
@@ -473,35 +473,54 @@ func (r *Reaper) GetObjectChangesMeasurement(ctx context.Context, md *sources.So
...
+		_, ok := availableExtentions[extToCreate]
 		if !ok {
 			log.GetLogger(ctx).Errorf("[%s] Requested extension %s not available on instance, cannot try to create...", md, extToCreate)
 		} else {
 			sqlCreateExt := `create extension ` + extToCreate
-			_, err := QueryMeasurements(ctx, md, sqlCreateExt)
-			if err != nil {
-				log.GetLogger(ctx).Errorf("[%s] Failed to create extension %s (based on --try-create-listed-exts-if-missing input): %v", md, extToCreate, err)
+			execArgs := []any{}
+			if !md.IsPostgresSource() {
```
pashagolub, line 515: If it's not a postgres source, we don't want to create any extensions. It's impossible to create extension in pg_bouncer, for example

internal/reaper/database.go
gemy26: Thanks for review, 
Could you clarify the correct logging format? , Should I use source-aware logger like: 
```l.WithField("source", md.Name).Infof("Failed to get a list of available extensions: %v", err) ```
Or would you prefer a different approach?

internal/reaper/database.go
gemy26: it's for lookup when checking if extension exist since we check each extension in `extensionNames` against the available extensions.

Issue #1166, opened 2026-02-02
Monitor indexes bloat


0xgouda:
We should add a new metric and panels that visualize the ratio of bloat for db indexes, probably in the form `Top $n indexes by bloat ratio`.

pashagolub:
Don't we have that information under table overview panels?

0xgouda:
> Don't we have that information under table overview panels?

No, we don't. We only have `Index Sizes Treemap` in the pg dashboard, and nothing related to indexes in the prom dashboard.

pashagolub:
Should go to contrib

PR #1188, opened 2026-02-09
Add Index Bloat overview panel to Prometheus dashboards

Mohamed-Elfardy:
Fixes: #1166

This PR adds an Index Bloat Overview section to the Tables dashboard.

Introduces a new row for index bloat metrics
Adds a table panel showing top 10 bloated indexes by bloat ratio in `tables-overview-prometheus`

### DIFF
--- a/grafana/prometheus/v12/4-tables-overview-prometheus.json
+++ b/grafana/prometheus/v12/4-tables-overview-prometheus.json
@@ -678,6 +678,80 @@
       ],
       "title": "Maintenance Operation Times",
       "type": "timeseries"
+    },
+    {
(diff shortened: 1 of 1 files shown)


0xgouda:
Thanks for your contribution, but please ask before working on issues if it's available to save time for both of us.

1. The problem isn't a dashboarding problem; we don't have a metric that measures index bloat.
2. The problem isn't for prom dashboards only
3. There is no `index_used_size_b` field in the `index_stats` metric (you are using it in your promQL query)

Mohamed-Elfardy:
Thanks for the clarification, I will make sure to discuss first.

PR #1182, opened 2026-02-06
Tests for `internal/sinks/prometheus.go`, fixes #1099

abhijeetsharma200:
Helper + Cache Operations tests in this commit more test coming.

PR #1190, opened 2026-02-09
[*] optimize `PrometheusWriter.Collect` locking, fixes #1189

### LINKED ISSUE #1189: Performance: Optimize locking in `Prometheus` Sink (Remove O(N) allocation in `Collect`)
`PrometheusWriter.Collect` currently holds a global write lock while iterating over all monitored databases to pre-allocate empty maps for the next scrape cycle.

```go
// internal/sinks/prometheus.go
promw.Lock()
snapshot := promw.Cache
promw.Cache = make(PromMetricCache, len(snapshot))

// O(N) LOOP INSIDE CRITICAL SECTION
for dbUnique := range snapshot {
    promw.Cache[dbUnique] = make(map[string]metrics.MeasurementEnvelope)
}
promw.Unlock()
```
This creates a bottleneck as the number of monitored databases grows. The lock is held for `O(N)` time, blocking all incoming `Write` operations during every scrape interval. This can cause latency spikes in metric ingestion.

Use lazy initialization to make Collect an `O(1)` operation:

- In `Collect`, remove the allocation loop and simply swap the cache with a new empty map.
- In `PromAsyncCacheAddMetricData`, initialize the inner map for a dbUnique only when data is written.

**Benefits**
- Constant-time locking in Collect
- Reduced write contention
- Avoids allocating maps for inactive databases

abhijeetsharma200:
Fixes #1189


pashagolub:
Thanks. It would be nice to have tests showing new functionality works as expected, even thought we didn't have tests previously

abhijeetsharma200:
> Thanks. It would be nice to have tests showing new functionality works as expected, even thought we didn't have tests previously

I think it would be better to merge #1182 first, and then I can move the test I wrote here into that file. So all future tests either go in `prometheus_test` or `prometheus_race_test`. Otherwise, prometheus would have too many testing files

pashagolub:
> I think it would be better to merge #1182 first, and then I can move the test I wrote here into that file. So all future tests either go in `prometheus_test` or `prometheus_race_test`. Otherwise, prometheus would have too many testing files

I don't see any problems. Rename the file to `prometheus_test.go` and put all available tests in it

abhijeetsharma200:
> > I think it would be better to merge #1182 first, and then I can move the test I wrote here into that file. So all future tests either go in `prometheus_test` or `prometheus_race_test`. Otherwise, prometheus would have too many testing files
> 
> I don't see any problems. Rename the file to `prometheus_test.go` and put all available tests in it

Won't that cause a merge conflict?

PR #1192, opened 2026-02-10
[-] fix file leak in `GetPathUnderlyingDeviceID`, fixes #1185

### LINKED ISSUE #1185: BUG: Resource leak in `GetPathUnderlyingDeviceID` (unclosed files)
**Describe the bug**
When trying to make tests to increase coverage I found that the function `GetPathUnderlyingDeviceID` in file `internal/reaper/psutil_linux.go` opens a file but never closes it. Which causes a file descriptor leak.

**To Reproduce**
Run below test:

``` go

func TestGetPathUnderlyingDeviceID_LeakCheck(t *testing.T) {
	// 1. Create a dummy file
	f, err := os.CreateTemp("", "leak_test")
	if err != nil {
		t.Fatal(err)
	}
	defer os.Remove(f.Name())
	f.Close()

	// 2. Count open files before the loop
	initialFDs, err := os.ReadDir("/proc/self/fd")
	if err != nil {
		t.Skip("Skipping test: cannot read /proc/self/fd (are you on Linux?)")
	}
	initialCount := len(initialFDs)

	// 3. Run the function 100 times
	for i := 0; i < 100; i++ {
		_, err := GetPathUnderlyingDeviceID(f.Name())
		if err != nil {
			t.Fatalf("Function failed: %v", err)
		}
	}

	// 4. Count open files after the loop
	finalFDs, err := os.ReadDir("/proc/self/fd")
	if err != nil {
		t.Fatal(err)
	}
	finalCount := len(finalFDs)

	// 5. Calculate the leak and log results
	diff := finalCount - initialCount

	t.Logf("Initial FDs: %d, Final FDs: %d, Leaked: %d", initialCount, finalCount, diff)

	if diff >= 100 {
		t.Errorf("FAIL: Resource leak detected! Leaked %d file descriptors in 100 iterations.", diff)
	} else {
		t.Log("PASS: No significant leak detected.")
	}
}
```

Output:
```
=== RUN   TestGetPathUnderlyingDeviceID_LeakCheck
    psutil_test.go:159: Initial FDs: 8, Final FDs: 108, Leaked: 100
    psutil_test.go:162: FAIL: Resource leak detected! Leaked 100 file descriptors in 100 iterations.
--- FAIL: TestGetPathUnderlyingDeviceID_LeakCheck (0.04s)
FAIL
FAIL    github.com/cybertec-postgresql/pgwatch/v5/internal/reaper       0.053s
FAIL
```

**Expected behavior**
Expected above test to `PASS` which would mean no leak and all files opened are closed.

Mazen050:
Fixes #1185 

Adds file close to avoid leak and unclosed files when using the function.

### DIFF
--- a/internal/reaper/psutil_test.go
+++ b/internal/reaper/psutil_test.go
@@ -1,6 +1,7 @@
 package reaper
 
 import (
+	"runtime"
 	"slices"
 	"testing"
 
@@ -8,6 +9,7 @@ import (
 
 	"maps"
 	"os"
+
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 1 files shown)


Mazen050:
sure, I was working on another PR that adds tests for this function should I add it in this PR or in the other PR focused on adding tests?

Also, I like this approach, requiring tests for new features and bug fixes helps avoid coverage problems like the one we have right now. similar to how projects like pandas enforce testing as part of changes.

Note: The PR adds tests for multiple files including this function in the tests.

pashagolub:
Usually code and tests are merged together. Also, I won't accept PR that will have tests for different parts of the code. Pull request should touch only one area, so it's better not to create mega PRs with many different areas involved. Consider this: I'm not happy with some tiny detail and now you cannot merge all tests, because they are packed in one pull request

Mazen050:
Fair enough I will add tests here and will divide my PR to 2 or 3 PRs depending on files.

Mazen050:
Sorry about that. Added the import now.

PR #1199, opened 2026-02-11
[-]: fix race conditions in 'fileMetricReader', fixes #1198

### LINKED ISSUE #1198: BUG: Race condition in YAML metrics file operations causes data loss under concurrent access
**Describe the bug**
The  `internal/metrics/yaml.go` has a  race condition in its create-delete-update operations. When multiple goroutines concurrently create, update, or delete metrics/presets, some operations are lost.

**To Reproduce**

```GO
func TestConcurrentUpdates(t *testing.T) {
	ctx := context.Background()
	tempDir := t.TempDir()
	tempFile := filepath.Join(tempDir, "metrics.yaml")

	yamlrw, err := metrics.NewYAMLMetricReaderWriter(ctx, tempFile)
	assert.NoError(t, err)

	// Create initial empty metrics file
	initialMetrics := &metrics.Metrics{
		MetricDefs: make(map[string]metrics.Metric),
		PresetDefs: make(map[string]metrics.Preset),
	}
	err = yamlrw.WriteMetrics(initialMetrics)
	assert.NoError(t, err)

	// Number of concurrent updates
	numGoroutines := 10
	var wg sync.WaitGroup
	errorChan := make(chan error, numGoroutines)

	// Each goroutine will add a unique metric
	for i := 0; i < numGoroutines; i++ {
		wg.Add(1)
		go func(id int) {
			defer wg.Done()

			metricName := fmt.Sprintf("metric_%d", id)
			testMetric := metrics.Metric{
				Description: fmt.Sprintf("Test metric %d", id),
				SQLs: map[int]string{
					1: fmt.Sprintf("SELECT %d", id),
				},
			}

			// Simulate small random delay
			time.Sleep(time.Millisecond * time.Duration(id%3))

			err := yamlrw.UpdateMetric(metricName, testMetric)
			if err != nil {
				errorChan <- fmt.Errorf("goroutine %d: %w", id, err)
			}
		}(i)
	}

	wg.Wait()
	close(errorChan)

	// Check for errors during updates
	var errors []error
	for err := range errorChan {
		errors = append(errors, err)
		t.Logf("Error during concurrent update: %v", err)
	}

	// ensure all metrics were saved
	finalMetrics, err := yamlrw.GetMetrics()
	assert.NoError(t, err)

	assert.Equal(t, numGoroutines, len(finalMetrics.MetricDefs),
		"Expected %d metrics, but got %d. Some updates were lost due to race condition!",
		numGoroutines, len(finalMetrics.MetricDefs))

	// Print which metrics are missing if test fails
	if len(finalMetrics.MetricDefs) != numGoroutines {
		for i := 0; i < numGoroutines; i++ {
			metricName := fmt.Sprintf("metric_%d", i)
			if _, exists := finalMetrics.MetricDefs[metricName]; !exists {
				t.Logf("LOST UPDATE: %s was not saved", metricName)
			}
		}
	}
}
```

**Expected behavior**
- All 10 concurrent metric updates should be saved
- No race conditions 
- No  data loss under concurrent access

**Screenshots**

Image

gemy26:
This PR fixes #1198 
- Add sync.Mutex to fileMetricReader struct.
- used mutex locks with all operations (Create/Update/Delete). 
- Add concurrent test cases.

### DIFF
--- a/internal/sources/yaml.go
+++ b/internal/sources/yaml.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"sync"
 
 	"gopkg.in/yaml.v3"
 )
@@ -23,6 +24,7 @@ func NewYAMLSourcesReaderWriter(ctx context.Context, path string) (ReaderWriter,
 type fileSourcesReaderWriter struct {
 	ctx  context.Context
 	path string
+	mu   sync.RWMutex
 }
 
 func (fcr *fileSourcesReaderWriter) WriteSources(mds Sources) error {
@@ -31,7 +33,9 @@ func (fcr *fileSourcesReaderWriter) WriteSources(mds Sources) error {
 }
 
 func (fcr *fileSourcesReaderWriter) UpdateSource(md Source) error {
-	dbs, err := fcr.GetSources()
+	fcr.mu.Lock()
+	defer fcr.mu.Unlock()
+	dbs, err := fcr.getSourcesNoLock()
 	if err != nil {
 		return err
 	}
@@ -46,7 +50,9 @@ func (fcr *fileSourcesReaderWriter) UpdateSource(md Source) error {
 }
 
 func (fcr *fileSourcesReaderWriter) CreateSource(md Source) error {
-	dbs, err := fcr.GetSources()
+	fcr.mu.Lock()
+	defer fcr.mu.Unlock()
+	dbs, err := fcr.getSourcesNoLock()
 	if err != nil {
 		return err
 	}
@@ -61,7 +67,9 @@ func (fcr *fileSourcesReaderWriter) CreateSource(md Source) error {
 }
 
 func (fcr *fileSourcesReaderWriter) DeleteSource(name string) error {
-	dbs, err := fcr.GetSources()
+	fcr.mu.Lock()
+	defer fcr.mu.Unlock()
+	dbs, err := fcr.getSourcesNoLock()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


0xgouda (changes requested):
No need to duplicate `mu.Lock()` `mu.Unlock()` everywhere we only need it around the critical section in the shared `fmr.WriteMetrics()` 

Also, we should use `RWLock` instead, and lock the critical section in `getMetrics()` while reading.

Also, please do the same for yaml sources handler

gemy26:
Thanks for reply, 
If we only lock inside the read and write functions, we can still have a race condition. in case of multiple goroutines may call CreateMetric (or any other operations) at the same time Each goroutine would : ```GetMetrics() -> modify -> WriteMetrics()```. Even if GetMetrics and WriteMetrics are protected, the sequence is not atomic. Two goroutines could read the same initial state, apply different changes, and then one write would overwrite the other, causing data loss.
Did i miss something ?

0xgouda:
Oh yeah, you are right, sorry for that, I got a bit confused.

But we still should use RWLocks.

gemy26:
I applied the changes and fixed sources yaml also I implemented getMetricsNoLock/getSourcesNoLock to be able to use RWLock and avoid deadlock in write operations.

Without it: CreateMetric acquires Lock -> Calls GetMetrics() which tries to acquire RLock() = Deadlock.

Please let me know if there's a better approach.

internal/sources/yaml.go
```diff
@@ -61,15 +67,17 @@ func (fcr *fileSourcesReaderWriter) CreateSource(md Source) error {
...
+	dbs, err := fcr.getSourcesNoLock()
 	if err != nil {
 		return err
 	}
 	dbs = slices.DeleteFunc(dbs, func(md Source) bool { return md.Name == name })
 	return fcr.WriteSources(dbs)
 }
 
-func (fcr *fileSourcesReaderWriter) GetSources() (dbs Sources, err error) {
+func (fcr *fileSourcesReaderWriter) getSourcesNoLock() (dbs Sources, err error) {
```
pashagolub, line 80: I don't like the name

gemy26: What about renaming to `readMetrics`/ `readSources`?

pashagolub: Why do you even touch it?

pashagolub:
Thanks a lot for your contribution! This PR was overtaken by #1209

PR #1216, opened 2026-02-20
fix: use defer to prevent resource leaks in convert_metrics tests

Makeepan-dev:
## Description
This PR resolves a resource leak identified in `cmd/convert_metrics/convert_metrics_test.go`. 

The original code manually closed file handles at the end of a block. If a failure occurred during execution (specifically at `t.Fatalf`), the handles for the destination file and the zip source file would remain open.

## Changes
- Implemented `defer out.Close()` and `defer rc.Close()` immediately after successful resource acquisition.
- This ensures all file descriptors are released correctly even in the event of a test failure or panic.

*Verified via static analysis auditing (MakeepanScanner).*


pashagolub:
Not a bug. 

PS Please stop using automatic scanners to file issues and propose PRs.

PR #1218, opened 2026-02-21
[*] remove `fmt.Println` from `webserver` tests

Makeepan-dev:
## Description
While reviewing the test suite, I noticed `fmt.Println` being used in `internal/webserver/server_test.go`. This pollutes the standard console output during test runs.

This PR replaces those instances with `t.Log()`, ensuring that test logs are only printed when a test fails or when the test suite is explicitly run with the verbose flag (`-v`). 

## Changes Made
* Replaced `fmt.Println` with `t.Log` in `server_test.go` to conform to standard Go testing practices and keep CI logs clean.

### DIFF
--- a/internal/webserver/server_test.go
+++ b/internal/webserver/server_test.go
@@ -2,7 +2,6 @@ package webserver_test
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


pashagolub:
Thanks. Please fix the issues

PR #1223, opened 2026-02-22
[-] resolve `monitoredSources` race condition, fixes #1213

### LINKED ISSUE #1213: Bug: Data race in `reaper.monitoredSources` causing potential crash or inconsistent reads
**Describe the bug**
There is a data race in `internal/reaper/reaper.go`. The `r.monitoredSources` slice is written to by LoadSources() (which runs in the main Reaper loop) while being read concurrently by WriteMonitoredSources() (which runs in a separate background goroutine).

**To Reproduce**
By running `go test -race ./...`

or to force it run this test with `-race`:
``` go
package reaper

import (
	"context"
	"fmt"
	"sync"
	"testing"
	"time"

	"github.com/cybertec-postgresql/pgwatch/v5/internal/cmdopts"
	"github.com/cybertec-postgresql/pgwatch/v5/internal/metrics"
	"github.com/cybertec-postgresql/pgwatch/v5/internal/sinks"
	"github.com/cybertec-postgresql/pgwatch/v5/internal/sources"
)

// --- Mocks ---

type MockSourceReader struct {
	sources []sources.Source
	mu      sync.Mutex
}

func (m *MockSourceReader) GetSources() (sources.Sources, error) {
	m.mu.Lock()
	defer m.mu.Unlock()
	ret := make(sources.Sources, len(m.sources))
	copy(ret, m.sources)
	return ret, nil
}

func (m *MockSourceReader) CreateSource(_ sources.Source) error        { return nil }
func (m *MockSourceReader) UpdateSource(_ sources.Source) error        { return nil }
func (m *MockSourceReader) DeleteSource(_ string) error                { return nil }
func (m *MockSourceReader) GetSource(_ string) (sources.Source, error) { return sources.Source{}, nil }
func (m *MockSourceReader) WriteSources(_ sources.Sources) error       { return nil }

type MockSinkWriter struct{}

func (m *MockSinkWriter) Write(msg metrics.MeasurementEnvelope) error                { return nil }
func (m *MockSinkWriter) SyncMetric(dbUnique, metric string, op sinks.SyncOp) error { return nil }
func (m *MockSinkWriter) Flush()                                                    {}


func TestReaper_Race_Condition_Crash(t *testing.T) {
	
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	opts := &cmdopts.Options{}
	opts.Sources.Refresh = 1

	r := NewReaper(ctx, opts)
	
	mockReader := &MockSourceReader{}
	r.SourcesReaderWriter = mockReader
	r.SinksWriter = &MockSinkWriter{}

	go r.WriteMeasurements(ctx)

	var wg sync.WaitGroup
	wg.Add(2)

	// 1. READER GOROUTINE
	go func() {
		defer wg.Done()
		for {
			select {
			case <-ctx.Done():
				return
			default:
				r.WriteMonitoredSources(ctx)
			}
		}
	}()

	// 2. WRITER GOROUTINE
	go func() {
		defer wg.Done()

		hugeList := make([]sources.Source, 10000)
		for i := 0; i < 10000; i++ {
			hugeList[i] = sources.Source{Name: fmt.Sprintf("db_%d", i), Kind: "postgres"}
		}
		emptyList := []sources.Source{}

		ticker := time.NewTicker(time.Microsecond * 100)
		defer ticker.Stop()
		toggle := false

		for {
			select {
			case <-ctx.Done():
				return
			case <-ticker.C:
				mockReader.mu.Lock()
				if toggle {
					mockReader.sources = hugeList
				} else {
					mockReader.sources = emptyList
				}
				toggle = !toggle
				mockReader.mu.Unlock()

				_ = r.LoadSources(ctx)
			}
		}
	}()

	time.Sleep(5 * time.Second)
	t.Log("Test finished.")
}
```
Output:
```
--- FAIL: TestReaper_Race_Condition_Crash (5.00s)
    reaperRace_test.go:110: Test finished.
    testing.go:1712: race detected during execution of test
2026-02-18 02:48:09.340 [INFO] [sources:0] sources refreshed
```

**Expected behavior**
Concurrent access to `r.monitoredSources` should be synchronized. Updates in `LoadSources` should be atomic and safe from concurrent reads in `WriteMonitoredSources`.

Mazen050:
Fix a data race in Reaper caused by concurrent access to the shared `monitoredSources` slice.

`LoadSources()` updates this slice while background goroutines (`Reap()` loop and `WriteMonitoredSources()`) read it concurrently, which triggers race detector warnings.

This PR protects the slice with an RWMutex and clones it before iteration to ensure safe concurrent access.

Fixes #1213

### DIFF
--- a/internal/reaper/reaper.go
+++ b/internal/reaper/reaper.go
@@ -39,7 +39,7 @@ type Reaper struct {
 	measurementCh        chan metrics.MeasurementEnvelope
 	measurementCache     *InstanceMetricCache
 	logger               log.Logger
-	mu 					 sync.RWMutex
+	mu                   sync.RWMutex
 	monitoredSources     sources.SourceConns
 	prevLoopMonitoredDBs sources.SourceConns
 	cancelFuncs          map[string]context.CancelFunc


internal/webserver/wslog_test.go
pashagolub, line 1: I'm not sure this file is somehow related to the fix

Mazen050: Sorry that's my bad, I forgot to mention that this specific test failed in CI after the fix, Also it didn't fail on my machine so its related to github's CI.

Mazen050:
> It would be perfect if you can add some tests

Sure thing, I was actually planning on adding tests for this specific file and had some tests already written. Will add these here asap.

internal/webserver/wslog_test.go
Mazen050: I attempted to fix it so that when you review we can see if my fix is good or we should take another approach.

Mazen050:
Hello @pashagolub 

I have added tests increasing coverage by about 9%. Only two functions remaining with 0% coverage `reapMetricMeasurements` and `Reap` which both I think are hard. please let me know if we should make tests for them so that I can try again with them and either send another PR or add to this. Also if we were to make a test for `Reap` how would we go about it?

Also please let me know if there are any edge cases we can add tests for that are not necessarily related to coverage.

pashagolub:
> @pashagolub, I wonder if, instead of adding additional locking overhead, we should deprecate `configured_dbs` and remove the `writeMonitoredSources()` function entirely. Almost all dashboards now rely on `admin.all_distinct_dbname_metrics` instead?

We use it in Global Health dashboards. If we can get rid of it, I will support this decision

Mazen050:
> We use it in Global Health dashboards. If we can get rid of it, I will support this decision

Hello I have investigated this and would just replacing 

``` sql 
(select dbname, max(time) from configured_dbs where time between now() - '15min'::interval and now() group by dbname) c
``` 

in the global health dashboard with `admin.all_distinct_dbname_metrics c` work? or Day-0 failures are needed here and `admin.all_distinct_dbname_metrics` doesn't provide this?

internal/reaper/reaper.go
```diff
@@ -37,6 +39,7 @@ type Reaper struct {
 	measurementCh        chan metrics.MeasurementEnvelope
 	measurementCache     *InstanceMetricCache
 	logger               log.Logger
+	mu 					 sync.RWMutex
```
pashagolub: tabs are used for indents and spaces are used for alignment

pashagolub:
> 
> ```sql
> (select dbname, max(time) from configured_dbs where time between now() - '15min'::interval and now() group by dbname) c
> ```
> 
> in the global health dashboard with `admin.all_distinct_dbname_metrics c` work?

Should work from what I see

Mazen050:
> Should work from what I see

@pashagolub Should I do it? and remove the `writeMonitoredSources()` function?

pashagolub:
closed by #1262

PR #1224, opened 2026-02-22
[-] return the error from `MultiWriter.DefineMetrics()`, fixes #1222

### LINKED ISSUE #1222: BUG: `MultiWriter.DefineMetrics` always returns `nil`, silently discarding errors from child writers
`MultiWriter.DefineMetrics` correctly accumulates errors from child writers using errors.Join, but then unconditionally returns nil instead of the accumulated error, silently discarding all failures

Mahmoud-Ayman-Saleh:
Fix: return the accumulated error instead of nil
Add TestDefineMetrics to cover the failure scenarios:
  - single definer returning an error
  - multiple definers both returning errors
  - mixed writers where one definer fails

### DIFF
--- a/internal/sinks/multiwriter_test.go
+++ b/internal/sinks/multiwriter_test.go
@@ -9,14 +9,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-type MockWriter struct{}
-
-func (mw *MockWriter) SyncMetric(_, _ string, _ sinks.SyncOp) error {
-	return nil
+// mockWriter implements Writer and Migrator interfaces
+type mockWriter struct {
+	err               error
+	needsMigration    bool
+	needsMigrationErr error
 }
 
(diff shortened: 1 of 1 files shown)


internal/sinks/multiwriter_test.go
```diff
@@ -179,6 +179,89 @@ func TestMultiWriterMigrate(t *testing.T) {
...
+		{
+			name: "two definers — both error (BUG: returns nil instead of joined error)",
+			writers: []sinks.Writer{
+				&mockMetricsDefiner{defineErr: assert.AnError},
+				&mockMetricsDefiner{defineErr: assert.AnError},
+			},
+			expectError: true,
+		},
+		{
+			name: "mixed writers — one error (BUG: returns nil instead of error)",
```
Mazen050: I think test names like this `(BUG: returns nil instead of error)` would be a bit confusing after fixing the bug. you should probably change these names.

PR #1226, opened 2026-02-22
[-] return HTTP 405 for non-POST methods in login handler

gemy26:
Added default case for unsupported methods.

### DIFF
--- a/internal/webserver/jwt.go
+++ b/internal/webserver/jwt.go
@@ -49,7 +49,7 @@ func (s *WebUIServer) handleLogin(w http.ResponseWriter, r *http.Request) {
 
 	default:
 		w.Header().Set("Allow", "POST")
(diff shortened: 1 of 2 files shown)


internal/webserver/jwt.go
```diff
@@ -48,8 +47,9 @@ func (s *WebUIServer) handleLogin(w http.ResponseWriter, r *http.Request) {
 		}
 		_, err = w.Write([]byte(token))
 
-	case "GET":
-		fmt.Fprintf(w, "only POST methods is allowed.")
+	default:
+		w.Header().Set("Allow", "POST")
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
```
Mazen050: You should probably match the error message with the old one or change the expected message in the test.

I think matching the error messages is best here. This will also fix the CI.

PR #1232, opened 2026-02-23
[-] fix `expandEnvVars()` to expand `connStr` and `Group` fields

0xgouda:
Relevant discussion: https://github.com/cybertec-postgresql/pgwatch/discussions/1230


pashagolub (approved):
Perfect!

PR #1239, opened 2026-02-24
[*] Fix `global-db-overview` query logic, fixes #1152

### LINKED ISSUE #1152: Problem in query logic of some panels
This pattern of using a column and its lag to calculate the individual diffs in the specified time range is used in many panels in several dashboards: 

```sql
SELECT
  $__timeGroup(time, $agg_interval),
  avg( ((seq_scan-seq_scan_lag)::numeric*3600) / extract(epoch from time - time_lag) ) as seq_scan,
  avg( ((idx_scan-idx_scan_lag)::numeric*3600) / extract(epoch from time - time_lag) ) as idx_scan
FROM (  
  SELECT
    (data->>'seq_scan')::int8 as seq_scan, lag((data->>'seq_scan')::int8) over w as seq_scan_lag,
    (data->>'idx_scan')::int8 as idx_scan, lag((data->>'idx_scan')::int8) over w as idx_scan_lag,
    time, lag(time) over w as time_lag  
  FROM
    table_stats
  WHERE
    $__timeFilter(time) AND dbname = '$dbname'
    AND tag_data @> jsonb_build_object('table_full_name', '$table_full_name')
  WINDOW w as (order by time)
) x
WHERE seq_scan >= seq_scan_lag and time > time_lag
GROUP BY 1
ORDER BY 1
```

The thing is, the `lag()` of the very first row is always null due to using `$__timeFilter(time)`, so `column - column_lag` also evaluates to null, hence skipping that row's value, making the results less precise.

using `$__timeFilter(time)` in the inner query (current approach):

Image

same panel, time interval, and everything but using `$__timeFilter(time)` in the outer query to avoid having the `lag()` of the first in-range row be null:

Image

Mazen050:
This PR partially fixes #1152 and fixes #1314 

Fixing query logic in the `global-db-overview` dashboard.


pashagolub:
closed by #1262

Mazen050:
@pashagolub this PR has nothing to do with my PR. My PR fixes another bug

pashagolub:
@Mazen050 sorry about that!

Mazen050:
Before:

image



After:

image

0xgouda:
Please describe how you fixed the queries.

Mazen050:
Hello @0xgouda , Based on [your fix](https://github.com/cybertec-postgresql/pgwatch/issues/1152#issuecomment-3820641773) I fixed the queries my self first then sent to gemini 3 for review.

Here basically we get the most recent sample before the dashboards selected time range so that `lag()` has something to subtract from for the first row in the range. preventing the `NULL` we used to get because the inner query's `WHERE time >= $__timeFrom()` excludes any prior row for it to look back at.

A new addition to your CTE is `DISTINCT ON (dbname)` which is necessary here because in this dashboard the queries run on multiple dbs so `DISTINCT ON (dbname)` makes us select the most recent rows(before range) per db.

Please if anything else needs more clarification let me know.

Mazen050:
I rebased and now go tests are failing.

Mazen050:
before:
Screenshot 2026-04-02 231642

after:
Screenshot 2026-04-02 231711


[explaination of fix](https://github.com/cybertec-postgresql/pgwatch/issues/1314#issuecomment-4091266403)

0xgouda:
@Mazen050, is this ready for review? Are you still on it?

Mazen050:
@0xgouda Sorry I was just abit busy the past few days I was planning to continue here asap. the remaining things are fixing [this](https://github.com/cybertec-postgresql/pgwatch/issues/1152#issuecomment-3981359991) with coalesce and checking other panels. Should be ready soon.

PR #1249, opened 2026-02-25
[*] Fix name edit bug in webui and backend, fixes #1248

### LINKED ISSUE #1248: Changing a metric's name duplicates it permanently
**Describe the bug**
When changing a metrics name and pressing update metric a new metric is created but the old one is not deleted so its basically duplicated. Even after refresh its still there.

**To Reproduce**
1. Go to metrics tab in webui
2. Click on edit icon on a metric
3. change name and press update metric
4. See duplicate

**Expected behavior**
Old metric with old name should be deleted after new metric is created.

**Screenshots**

Image

changing name from `example1` to `example2` and pressing update metric gives the duplication below:

Image

Mazen050:
This PR fixes the bug where when we rename a metric/preset the old metric/preset is still there so it just creates a copy of it with another name. With this PR renaming works as intended.

Fixes #1248 

Note: Sorry for many PRs I just enjoyed this PR I couldn't keep it for later.

### DIFF
--- a/internal/metrics/default.go
+++ b/internal/metrics/default.go
@@ -30,7 +30,7 @@ func (dmrw *defaultMetricReader) DeleteMetric(string) error {
 	return errors.ErrUnsupported
 }
 
-func (dmrw *defaultMetricReader) UpdateMetric(string, Metric) error {
+func (dmrw *defaultMetricReader) UpdateMetric(string, string, Metric) error {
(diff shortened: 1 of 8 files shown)


0xgouda:
Hey, please be patient and dont overwhelme us with PRs, actually i dont think the issue you are referring to is an actual bug, so we should have a discussion first.

PR #1262, opened 2026-03-02
[*] deprecate `configured_dbs`

0xgouda:
- Deprecate `configured_dbs` dashboards, should rely on `admin.all_distinct_dbname_metrics` only from now on.
- Remove `WriteMonitoredSources()` function from reaper
- Update the `Global Health` pg dashboard to query `instance_up` for identifying offline nodes instead of `configured_dbs`
image

Closes: #1213

### DIFF
--- a/internal/reaper/metric.go
+++ b/internal/reaper/metric.go
@@ -11,11 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const (
-	monitoredDbsDatastoreSyncIntervalSeconds = 600              // write actively monitored DBs listing to metrics store after so many seconds
-	dbMetricJoinStr                          = "¤¤¤"            // just some unlikely string for a DB name to avoid using maps of maps for DB+metric data
-
-)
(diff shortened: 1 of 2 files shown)


pashagolub:
> * should rely on `admin.all_distinct_dbname_metrics`

Did you mean `instance_up`?

0xgouda:
> Did you mean `instance_up`?

In this specific dashboard, yeah use `instance_up` instead, but I meant if someone were to use it as way to list active sources `admin.al_...` is the way to go

pashagolub:
image
One notice. If source cannot be found in measurements at all, e.g. never successfully connected, the cell will be empty and confusing. Should we use smth like "Too long ago" in case case?

0xgouda:
> One notice. If source cannot be found in measurements at all, e.g. never successfully connected, the cell will be empty and confusing. Should we use smth like "Too long ago" in case case?

I guess "Unknown" is better?

pashagolub:
yeah, uknown is good

PR #1264, opened 2026-03-02
[*] fix docs update grafana install steps and add yaml setup example

kevo-1:
1. The Grafana install step uses the deprecated `apt-key` command (removed in Ubuntu 22.04+) and points to the old `packages.grafana.com` URL.
2. The YAML setup section never shows how to actually pass the YAML file to pgwatch via `--sources`


pashagolub:
Hi, thanks! Would you please provide a link with Grafana installation instructions?

kevo-1:
Sure! The steps are based on the official Grafana documentation for Debian/Ubuntu:
https://grafana.com/docs/grafana/latest/setup-grafana/installation/debian/

After reviewing it more closely, I noticed my snippet is slightly off from what the official docs currently recommend:
- The keyring path should be `/etc/apt/keyrings/grafana.asc` instead of `/usr/share/keyrings/grafana.gpg`
- The official docs use `gpg-full.key` fetched directly via `wget -O` rather than piping through `gpg --dearmor`

Here's the updated snippet that matches the official docs exactly:

```bash
sudo mkdir -p /etc/apt/keyrings
sudo wget -O /etc/apt/keyrings/grafana.asc https://apt.grafana.com/gpg-full.key
sudo chmod 644 /etc/apt/keyrings/grafana.asc

echo "deb [signed-by=/etc/apt/keyrings/grafana.asc] https://apt.grafana.com stable main" | \
sudo tee /etc/apt/sources.list.d/grafana.list

sudo apt-get update && sudo apt-get install grafana
```

I'll update the PR with this. Let me know if you'd like any other changes!

pashagolub:
Thanks

PR #1271, opened 2026-03-03
Expand `relkinds` in `buffercache_by_type` metric

nadashaban11:
## Description

I have noticed the `TODO` in the `buffercache_by_type` metric, then I explored the official doc of PostgreSQL to expand 4 other classifications.
## Fixes 

Added the following cases based on the doc:
**S**  => "Sequence"
**v** =>  "View"
**p**  => "Partitioned table"
**I** => "Partitioned index" 

## Screenshot 
Screenshot from 2026-03-03 22-53-47

## AI & Automation Policy

AI helped me to know the query that gives me the `relkinds` in my DB

## Checklist

- [x] New or updated tests are included where applicable (Manually testing using `psql`).
- [x] Documentation is updated where applicable (`metrics.yaml`).


pashagolub:
Could you please explain what kind of buffer cache statistics have Sequence and View?

Did you check the metric query output on a real database?

Thanks

nadashaban11:
Sure,  when I searched and explored docs, I found that when we create Sequence, which is as a single-row table, then it has a storage and when `nextval()` called, stored Sequence is loaded to buffer. then it can affect and appear in cache usage. 
If there are many updates, Sequences can also cause I/O overhead, so specifying its kind may help monitoring accuracy.

The previous query result shown in (screenshot) is from my local running database. If you mean checking from  pgwatch itself, I did not but I can do that if you want. 

For standard View it does not actually occupy buffers, but I included it to cover all `relkind` types I found in PostgreSQL doc. If it does not make sense I will remove it.

pashagolub:
Why didn't you add 'Foreign table' for example?

nadashaban11:
I think it not stored in local files that postgres buffers manage. or am I misunderstanding that?

pashagolub:
so postgres buffers manage local files?

nadashaban11:
What I know is that postgres storage manager manages the movement of local files from local storage into buffers when these files are needed to be accessible by any operation like read, write, or update. So if we have file is not stored locally, we can not manage by postgres. 
Please correct me if I'm wrong.

pashagolub:
thanks for your contribution. overtaken by #1305

Issue #1273, opened 2026-03-04
Problem on migrating from 4.1.0 to 5.0.0


clmntpllr:
After installing the 5.0.0 rpm package (Rhel 8.10), I launched the command `pgwatch config upgrade` 
and got the error : 

```
2026-03-04 09:32:04.095 [ERROR] [callstack:goroutine 1 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.26.0/x64/src/runtime/debug/stack.go:26 +0x5e
main.main.func1()
        /home/runner/work/pgwatch/pgwatch/cmd/pgwatch/main.go:50 +0xc5
panic({0xed6520?, 0x2f92cf4a6d20?})
        /opt/hostedtoolcache/go/1.26.0/x64/src/runtime/panic.go:860 +0x13a
github.com/cybertec-postgresql/pgwatch/v5/internal/cmdopts.(*ConfigUpgradeCommand).Execute.func1({0x2f92cf05805b, 0x3f}, 0x2f92cf2c1b20)
        /home/runner/work/pgwatch/pgwatch/internal/cmdopts/cmdconfig.go:110 +0x159
github.com/cybertec-postgresql/pgwatch/v5/internal/cmdopts.(*ConfigUpgradeCommand).Execute(0x1bc7070?, {0xeadc80?, 0x7f96a4b2da01?, 0x7f96eb7e04e8?})
        /home/runner/work/pgwatch/pgwatch/internal/cmdopts/cmdconfig.go:114 +0xad
github.com/jessevdk/go-flags.(*Parser).ParseArgs(0x2f92cf3bc070, {0x2f92cf050550, 0x2, 0x2})
        /home/runner/go/pkg/mod/github.com/jessevdk/go-flags@v1.6.1/parser.go:339 +0xacc
github.com/jessevdk/go-flags.(*Parser).Parse(...)
        /home/runner/go/pkg/mod/github.com/jessevdk/go-flags@v1.6.1/parser.go:195
github.com/cybertec-postgresql/pgwatch/v5/internal/cmdopts.New({0x12106e0, 0x2f92cf014040})
        /home/runner/work/pgwatch/pgwatch/internal/cmdopts/cmdoptions.go:76 +0xd9
main.main()
        /home/runner/work/pgwatch/pgwatch/cmd/pgwatch/main.go:59 +0xd7
] interface conversion: *sources.dbSourcesReaderWriter is not db.Migrator: missing method Migrate
[...]

PR #1274, opened 2026-03-04
[-] fix sources handling in `config upgrade`

0xgouda:
- don't run migrations for sources on `config upgrade`
    - There aren't any migrations for sources and the pg sources reader/writer
doesn't even implement a `Migrate()` method, so it's not necessary, and calling `Migrate()` on it will trigger a crash.

- Update `config upgrade` help messages
    - They are now more clear and state which flag the user needs to specify
for this migration to happen correctly.

    - This is especially useful for users who specify `--sources` without `--metrics`,
then it will get used as the metrics database and on migration users will have a clue that
they have to explicitly specify `--metrics` for the upgrade to happen.

Closes: #1273

### DIFF
--- a/internal/cmdopts/cmdconfig_test.go
+++ b/internal/cmdopts/cmdconfig_test.go
@@ -341,7 +341,7 @@ func TestConfigUpgradeCommand_Execute_Coverage(t *testing.T) {
 		cmd := ConfigUpgradeCommand{owner: opts}
 		err := cmd.Execute(nil)
 		a.Error(err)
-		a.ErrorContains(err, "at least one of --sources, --metrics, or --sink must be specified")
+		a.ErrorContains(err, "at least --metrics or --sink must be specified")
 		a.Equal(ExitCodeConfigError, opts.ExitCode)
 	})
 
@@ -449,7 +449,7 @@ func TestConfigUpgradeCommand_Execute_Coverage(t *testing.T) {
 	t.Run("only sources specified as yaml - unsupported error", func(*testing.T) {
 		var output bytes.Buffer
 		opts := &Options{
-			Sources:      sources.CmdOpts{Sources: "/tmp/sources.yaml"},
+			Metrics:      metrics.CmdOpts{Metrics: "/tmp/sources.yaml"},
 			OutputWriter: &output,
 		}
 		cmd := ConfigUpgradeCommand{owner: opts}


internal/cmdopts/cmdconfig_test.go
```diff
@@ -449,7 +449,7 @@ func TestConfigUpgradeCommand_Execute_Coverage(t *testing.T) {
 	t.Run("only sources specified as yaml - unsupported error", func(*testing.T) {
 		var output bytes.Buffer
 		opts := &Options{
-			Sources:      sources.CmdOpts{Sources: "/tmp/sources.yaml"},
+			Metrics:      metrics.CmdOpts{Metrics: "/tmp/sources.yaml"},
```
Mazen050, line 452: Hey just a heads up. After the change this test becomes a duplicate of the previous test (`only metrics specified as yaml - unsupported error`)

internal/cmdopts/cmdconfig.go
```diff
@@ -111,11 +111,7 @@ func (cmd *ConfigUpgradeCommand) Execute([]string) (err error) {
 
 	}
 
-	err = f(opts.Sources.Sources, func() (any, error) {
-		return sources.NewPostgresSourcesReaderWriter(ctx, opts.Sources.Sources)
-	})
-
-	err = errors.Join(err, f(opts.Metrics.Metrics, func() (any, error) {
+	err = errors.Join(f(opts.Metrics.Metrics, func() (any, error) {
```
Mazen050, line 114: Also why not remove `errors.join` here? since we have one error now?

0xgouda: because otherwise the error won't implement `Unwrap()`

docs/reference/cli_env.md
```diff
@@ -204,9 +204,9 @@ It reads the configuration from the specified sources and metrics, then begins c
 
 - `upgrade`
 
-    Upgrade the configuration and/or sink database to the latest version by applying all pending migrations.
+    Upgrades the metrics configuration and/or sink database to the latest version by applying all pending migrations.
```
pashagolub, line 207: It also update sources configuration on Postgres. Because they are stored together

docs/reference/cli_env.md
```diff
@@ -204,9 +204,9 @@ It reads the configuration from the specified sources and metrics, then begins c
...
     File or folder based configurations are not supported. The command will automatically detect which
-    databases (sources, metrics, sinks) require migrations and apply them.
+    databases (metrics, sinks) require migrations and apply them.
```
pashagolub, line 209: It will update sources stored in Postgres.

pashagolub:
> * There aren't any migrations for sources and the pg sources reader/writer
>       doesn't even implement a `Migrate()` method

because metrics reader takes care of it because sources and metrics are stored in the same schema of the same database

docs/reference/cli_env.md
0xgouda: What if a user (for some weird reason) uses to separate dbs for sources and metrics?

0xgouda: ping @pashagolub

pashagolub: that's a good question. never tested in real life. I assume pgwatch will take --metrics and use it as a default database while ignoring the --sources silently

0xgouda: That's why I removed source-related docs and code. 

There haven't been any migrations for sources yet, so we can only ask for `--metrics` and `--sink`, and we don't care about sources for now. 

And in the future, when we wish to add source migrations (if ever), we can start thinking about how we will solve this problem.

pashagolub: I see your point. but sources schema can be migrated. and user must know it. the fact we do storage in the same database is our problem. if you remove source upgrade mentions, user will stuck.

pashagolub: the clear way may be to make sure source writer is responsible for creating it's own tables and functions.

0xgouda: Yeah, but I mean for now they are not upgradable. Once they become so in future releases, we can update docs/code and see how to handle this problem, its not urgent/necessary now.

pashagolub: no. docs must tell that sources schema is upgradable. it's not important that upgrade is done with metric writer. user must know that all main database schemas are upgradable: sources, metrics, sinks.

0xgouda:
Overtaken by: #1287

PR #1316, opened 2026-03-19
[!] reaper goroutine consolidation via `pgx.Batch` queries

pashagolub:
The improvements stem from consolidating N per-metric goroutines into a single per-source `SourceReaper` that uses a GCD-based (Greatest Common Divisor) tick loop and `pgx.Batch` for batched SQL execution, dramatically reducing goroutine overhead, connection contention, scheduling pressure, and memory allocation churn.

# Performance Comparison: Batch Metric Fetching vs Master

## Test Environment

| Parameter | Value |
|-----------|-------|
| **Date** | 2026-03-19 |
| **Host OS** | Windows, Docker Desktop |
| **PostgreSQL** | 18.1 (Debian) |
| **Sources** | 6 (`debug` preset — 55+ metrics at 30s intervals) |
| **Sink** | PostgreSQL |
| **Log Level** | debug |
| **Build Tags** | `pprof` (enables `/debug/pprof/` on port 6060) |
| **Test Duration** | ~8 minutes per implementation |

### Branches Tested

| Branch | Commit | Description |
|--------|--------|-------------|
| `master` | `393b9529a` | Baseline — one goroutine per metric per source |
| `batch_metric_fetching` | `3d70c78a4` | New — one SourceReaper goroutine per source, GCD-based tick loop with `pgx.Batch` |

---

## Architecture Comparison

| Aspect | Master | Batch |
|--------|--------|-------|
| **Metric collection model** | 1 goroutine per metric per source | 1 SourceReaper goroutine per source |
| **SQL execution model** | Individual queries, one per goroutine | Batched queries via `pgx.Batch` |
| **Scheduling** | Per-metric `time.After` in each goroutine | Single GCD-based tick loop per source |

---

### Total Goroutine Count

| Measurement | Master | Batch | Reduction |
|-------------|--------|-------|-----------|
| **Total goroutines** | **435** | **28** | **93.6% (15.5× fewer)** |

### Goroutine Breakdown by Component

| Component | Master | Batch | Notes |
|-----------|--------|-------|-------|
| `reapMetricMeasurements` | **414** | **0** | Replaced by SourceReaper |
| `SourceReaper.Run` | 0 | **6** | 1 per source (new batch model) |
| `pgxpool.backgroundHealthCheck` | 9 | 9 | Connection pool health (unchanged) |
| `log.BrokerHook.poll` | 2 | 2 | Log dispatch (unchanged) |
| `sinks.PostgresWriter` | 2 | 2 | Sink writer goroutines (unchanged) |
| `reaper.Reap` (main loop) | 1 | 1 | Unchanged |
| `reaper.WriteMeasurements` | 1 | 1 | Unchanged |
| `webserver.Init` | 1 | 1 | Web UI (unchanged) |
| `net/http` | 3 | 3 | HTTP servers (pprof + web) |
| `signal` | 1 | 1 | OS signal handler |
| Other/runtime | 1 | 2 | Runtime overhead |

The batch implementation eliminates **414 per-metric goroutines**, replacing them with just **6 per-source goroutines**. All infrastructure goroutines (pool health checks, sinks, logging, web server) remain identical between both implementations.

### Stack Memory

| Metric | Master | Batch | Reduction |
|--------|--------|-------|-----------|
| **Stack in use** | 6.0 – 6.3 MB | 1.4 – 1.6 MB | **~74%** |

The 4× reduction in stack memory directly reflects the goroutine count difference — each goroutine requires its own stack.

---

### Heap Allocation (`Alloc` — live objects on the heap)

| Time (relative) | Master (KB) | Batch (KB) | Reduction |
|-----------------|-------------|------------|-----------|
| t=0 (startup) | 2,147 | 2,327 | — |
| t=2min | 29,698 | 3,983 | **87%** |
| t=4min | 27,257 | 3,837 | **86%** |
| t=6min | 29,209 | 3,903 | **87%** |
| t=8min | 17,651 | 3,865 | **78%** |

### Cumulative Allocations (`TotalAlloc` — total bytes allocated over time)

| Time (relative) | Master (KB) | Batch (KB) | Reduction |
|-----------------|-------------|------------|-----------|
| t=0 (startup) | 3,326 | 3,328 | — |
| t=2min | 865,826 | 207,017 | **76%** |
| t=4min | 1,620,935 | 294,202 | **82%** |
| t=6min | 2,402,248 | 382,324 | **84%** |
| t=8min | 3,178,864 | 474,467 | **85%** |

### System Memory (`Sys` — total memory obtained from OS)

| Time (relative) | Master (KB) | Batch (KB) | Reduction |
|-----------------|-------------|------------|-----------|
| t=2min | 96,890 | 28,914 | **70%** |
| t=8min | 96,890 | 29,170 | **70%** |

---

## Heap Profile Analysis (via `pprof /debug/pprof/heap`)

| Metric | Master | Batch | Reduction |
|--------|--------|-------|-----------|
| **HeapAlloc** (live) | 35.9 – 39.7 MB | 5.4 – 5.7 MB | **~85%** |
| **HeapSys** (reserved from OS) | 78.1 – 80.0 MB | 18.6 – 18.9 MB | **~76%** |
| **HeapObjects** (live objects) | 422K – 510K | 45K – 47K | **~91%** |
| **MaxRSS** (peak RSS) | 99.6 MB | 35.0 MB | **65%** |
| **GCCPUFraction** | 0.019 – 0.022% | 0.006 – 0.008% | **~66%** |

---

## Docker Container Stats (at steady state)

| Metric | Master | Batch | Improvement |
|--------|--------|-------|-------------|
| **Container Memory** | 56.7 – 63.2 MiB | 16.4 MiB | **~73% (3.5–3.9× less)** |
| **CPU Usage** | 1.84% | 0.03% | **~98% (61× less)** |
| **PIDs** | 22 | 17 | 23% |

---

## Allocation Rate Analysis

Calculated from `TotalAlloc` deltas between measurements (each 2-minute window):

| Window | Master (KB/min) | Batch (KB/min) | Ratio |
|--------|-----------------|----------------|-------|
| 0→2min | 431,250 | 101,845 | **4.2×** less |
| 2→4min | 377,555 | 43,593 | **8.7×** less |
| 4→6min | 390,657 | 44,061 | **8.9×** less |
| 6→8min | 388,308 | 46,072 | **8.4×** less |

At steady state (after initial warmup), the batch implementation allocates approximately **8–9× less memory per minute**.

---

## GC Pressure Analysis

| Metric | Master | Batch | Improvement |
|--------|--------|-------|-------------|
| GC cycles in 8 min | 176 | 170 | Similar |
| GCCPUFraction | 0.019 – 0.022% | 0.006 – 0.008% | **~66% less** |
| Avg GC pause (from pprof) | Higher variance | Lower, more consistent | — |

While GC cycle counts are similar, the CPU fraction spent on GC is ~3× lower for batch because each cycle processes far fewer live objects (45K vs 510K).

---

## Summary

The `batch_metric_fetching` branch delivers significant resource improvements across all dimensions:

| Category | Improvement | Detail |
|----------|-------------|--------|
| **Goroutines** | **93.6% fewer** | 28 vs 435 (414 per-metric goroutines eliminated) |
| **Live heap memory** | **~85% less** | 5.5 MB vs 37 MB (HeapAlloc via pprof) |
| **Heap objects** | **~91% less** | 46K vs 466K live objects |
| **Stack memory** | **~74% less** | 1.5 MB vs 6.2 MB |
| **Peak RSS** | **65% less** | 35 MB vs 100 MB |
| **System memory** | **~70% less** | 29 MB vs 97 MB (Sys) |
| **Container memory** | **~73% less** | 16 MiB vs 60 MiB |
| **Allocation rate** | **~8–9× lower** | 45 MB/min vs 389 MB/min (steady state) |
| **CPU usage** | **~98% less** | 0.03% vs 1.84% |
| **GC CPU fraction** | **~66% less** | 0.007% vs 0.020% |

### DIFF
--- a/internal/reaper/source_reaper.go
+++ b/internal/reaper/source_reaper.go
@@ -145,6 +145,12 @@ func (sr *SourceReaper) Run(ctx context.Context) {
 			if lf := sr.lastFetch[name]; !lf.IsZero() && now.Sub(lf) < interval {
 				continue
 			}
+
+			metric, ok := metricDefs.GetMetricDef(name)
+			if !ok || sr.isRoleExcluded(metric) {
+				continue
+			}
+
 			switch {
 			case name == specialMetricServerLogEventCounts:
 				if sr.lastFetch[name].IsZero() {
@@ -157,16 +163,8 @@ func (sr *SourceReaper) Run(ctx context.Context) {
 			case IsDirectlyFetchableMetric(sr.md, name):
 				err = sr.fetchOSMetric(ctx, name)
 			case name == specialMetricChangeEvents || name == specialMetricInstanceUp:
-				err = sr.fetchSpecialMetric(ctx, name)
+				err = sr.fetchSpecialMetric(ctx, name, metric.StorageName)
 			default:
-				metric, ok := metricDefs.GetMetricDef(name)
-				if !ok {
-					l.WithField("metric", name).Warning("metric definition not found")
-					continue
-				}
-				if sr.isRoleExcluded(metric) {
-					continue
-				}
 				if cached := sr.reaper.GetMeasurementCache(sr.cacheKey(metric, name)); len(cached) > 0 {
 					l.WithField("metric", name).Info("instance level cache hit")
 					sr.sendEnvelope(ctx, name, metric.StorageName, cached)
@@ -264,14 +262,7 @@ func (sr *SourceReaper) fetchOSMetric(ctx context.Context, name string) error {
 }
 
 // fetchSpecialMetric handles change_events and instance_up metrics.
-func (sr *SourceReaper) fetchSpecialMetric(ctx context.Context, name string) error {
-	metric, ok := metricDefs.GetMetricDef(name)
-	if !ok {
-		return fmt.Errorf("metric definition not found for %s", name)
-	}
-	if sr.isRoleExcluded(metric) {
-		return nil
-	}
+func (sr *SourceReaper) fetchSpecialMetric(ctx context.Context, name, storageName string) error {
 	var (
 		data metrics.Measurements
 		err  error
@@ -286,7 +277,7 @@ func (sr *SourceReaper) fetchSpecialMetric(ctx context.Context, name string) err
 		return fmt.Errorf("failed to fetch special metric: %v", err)
 	}
 	if len(data) > 0 {
-		sr.sendEnvelope(ctx, name, metric.StorageName, data)
+		sr.sendEnvelope(ctx, name, storageName, data)
 	}
 	return err
 }
--- a/internal/reaper/source_reaper.go
+++ b/internal/reaper/source_reaper.go
@@ -20,18 +20,20 @@ const minTickInterval = 1 // seconds - floor for GCD to help handle zero/negativ
 // and batches SQL queries via pgx.Batch when the source is a real Postgres
 // connection (non-pgbouncer, non-pgpool).
 type SourceReaper struct {
-	reaper      *Reaper
-	md          *sources.SourceConn
-	lastFetch   map[string]time.Time
-	lastUptimeS int64 // last seen postmaster_uptime_s for restart detection
+	reaper          *Reaper
+	md              *sources.SourceConn
+	lastFetch       map[string]time.Time
+	lastUptimeS     int64               // last seen postmaster_uptime_s for restart detection
+	degradedMetrics map[string]struct{} // metrics that failed individual retry; executed via fetchMetric until they recover
 }
 
 // NewSourceReaper creates a SourceReaper for the given source connection.
 func NewSourceReaper(r *Reaper, md *sources.SourceConn) *SourceReaper {
 	sr := &SourceReaper{
-		reaper:    r,
-		md:        md,
-		lastFetch: make(map[string]time.Time),
+		reaper:          r,
+		md:              md,
+		lastFetch:       make(map[string]time.Time),
+		degradedMetrics: make(map[string]struct{}),
 	}
 	return sr
 }
@@ -179,6 +181,16 @@ func (sr *SourceReaper) Run(ctx context.Context) {
 					sr.lastFetch[name] = time.Now()
 					break
 				}
+				if _, degraded := sr.degradedMetrics[name]; degraded {
+					if err = sr.fetchMetric(ctx, batchEntry{name: name, metric: metric, sql: sql}); err != nil {
+						l.WithError(err).WithField("metric", name).Error("degraded metric fetch failed")
+					} else {
+						l.WithField("metric", name).Info("degraded metric recovered, returning to batch execution")
+						delete(sr.degradedMetrics, name)
+					}
+					sr.lastFetch[name] = time.Now()
+					break
+				}
 				batch = append(batch, batchEntry{name: name, metric: metric, sql: sql})
 			}
 			if err != nil {
@@ -212,8 +224,12 @@ func (sr *SourceReaper) Run(ctx context.Context) {
 }
 
 // executeBatch sends all SQLs in a single pgx.Batch round-trip, dispatching
-// each result immediately as it arrives. Individual query failures produce nil
-// data and are aggregated into the returned error.
+// each result immediately as it arrives. If any query fails, PostgreSQL's
+// extended protocol aborts all subsequent queries in the same sync boundary
+// (cascade failure). Any entry that returns an error from the batch is retried
+// individually via fetchMetric to isolate real failures from cascade failures.
+// Entries that fail even after the individual retry are marked as degraded
+// so that subsequent runs use fetchMetric for them until they recover.
 func (sr *SourceReaper) executeBatch(ctx context.Context, entries []batchEntry) error {
 	batch := &pgx.Batch{}
 	for _, e := range entries {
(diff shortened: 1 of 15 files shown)


internal/reaper/source_reaper.go
```diff
@@ -0,0 +1,327 @@
...
+// activeMetrics returns a snapshot copy of the currently active metric intervals
+// based on the source's recovery state. Copying under the lock prevents data
+// races when the caller iterates after the lock is released.
+func (sr *SourceReaper) activeMetrics() map[string]time.Duration {
+	sr.md.RLock()
+	defer sr.md.RUnlock()
+	src := sr.md.Metrics
+	if sr.md.IsInRecovery && len(sr.md.MetricsStandby) > 0 {
+		src = sr.md.MetricsStandby
+	}
```
0xgouda, line 50: The usage of `RLock()` is confusing me, not that we shouldn't be using it, but are we using `Lock()` and `RLock()` in other places that update/read from `md.Metrics`? From what I can see we don't.

internal/reaper/source_reaper.go
```diff
@@ -0,0 +1,327 @@
...
+					l.WithField("source", sr.md.Name).WithField("version", sr.md.Version).Warning("no SQL found for metric version")
+					break
+				}
+				batch = append(batch, batchEntry{name: name, metric: metric, sql: sql})
+			}
+			if err != nil {
+				l.WithError(err).WithField("metric", name).Error("failed to fetch metric")
+			}
+			sr.lastFetch[name] = now
+		}
```
0xgouda, line 204: There is an inconsistency here, for the special metrics the `lastFetch` map is updated after the metric fetch but for regular metrics its updated directly after batching the query.

The old approach for doing this was that for all metrics the sleep interval is started (corresponding to updating `lastFetch` here) after the metric is fetched to include the metric query processing time.

internal/reaper/source_reaper.go
```diff
@@ -0,0 +1,327 @@
...
+	"fmt"
+	"time"
+
+	"github.com/cybertec-postgresql/pgwatch/v5/internal/log"
+	"github.com/cybertec-postgresql/pgwatch/v5/internal/metrics"
+	"github.com/cybertec-postgresql/pgwatch/v5/internal/sources"
+	"github.com/jackc/pgx/v5"
+)
+
+const minTickInterval = 5 // seconds – floor for GCD to prevent excessive wake-ups
```
0xgouda: Ummm, this will set the minimum fetching interval to `5` even if specified otherwise in the configs.

Maybe this should be `1` instead?

internal/reaper/source_reaper.go
```diff
@@ -0,0 +1,330 @@
...
+				if sr.lastFetch[name].IsZero() {
+					go func() {
+						if e := sr.runLogParser(ctx); e != nil {
+							l.WithError(e).Error("log parser error")
+						}
+					}()
+				}
+			case IsDirectlyFetchableMetric(sr.md, name):
+				err = sr.fetchOSMetric(ctx, name)
+			case name == specialMetricChangeEvents || name == specialMetricInstanceUp:
```
0xgouda, line 170: Why aren't we checking `isRoleExcluded(metric)` for the first 2 cases? This will be buggy if a user modifies the `server_log_event_counts` or `psutil_*` metrics to have a specific required `node_status`

0xgouda: Fixed in: 4256426

internal/sources/conn.go
```diff
@@ -178,7 +178,7 @@ func (md *SourceConn) FetchRuntimeInfo(ctx context.Context, forceRefetch bool) (
 		return ctx.Err()
 	}
 
-	if !forceRefetch && md.LastCheckedOn.After(time.Now().Add(time.Minute*-2)) { // use cached version for 2 min
+	if !forceRefetch && md.LastCheckedOn.After(time.Now().Add(time.Minute*-5)) { // use cached version for 2 min
```
Mazen050:
```suggestion
	if !forceRefetch && md.LastCheckedOn.After(time.Now().Add(time.Minute*-5)) { // use cached version for 5 min
```
Cache duration looks like it was updated to 5 minutes, updating comment to reflect that

internal/reaper/source_reaper.go
0xgouda: Updated it to be 1 in 077cfd9

pashagolub: This was set to 5 for a reason but ok, let's try 1 and see if somebody will burn their prod :)

0xgouda: Ummm, I don't think the **worst-case** scenario where the gcd calculation falls back to 1, and the source go routine wakes up every second, finds nothing to do, gets back to sleep instantaneously, is going to burn production deployments?

internal/reaper/source_reaper.go
0xgouda: 1. Fixed this specific case for `md.Metrics` and `md.MetricsStandby` in d26c03d, so now the `LoadMetrics()` will always grab `md.Lock()` before updating these structs, and given that the source reaper always uses `md.RLock()` before accessing them, we are safe.

2. An important note: there are other places in the code that read from `md.Metrics` and `md.MetricsStandby`, but all of them execute in the main reaper routine - which is the only routine that issues a write in `LoadMetrics()` as stated above - so I saw that there is no need to grab additional `md.RLock()`[s] for them, as the sequential order guarantees safety.

3. Also, I noticed that the source `RWMutex` is rarely used to protect fields on write to them, I doubt if we do have race conditions because of that. This needs careful investigation in a separate issue.

@pashagolub I need your review on this. Specifically, what do you think of point 2 in terms of maintainability? Should we explicitly use `md.RLock()` so everything is extra clear and hence avoid introducing buggy changes in future updates/PRs? Or document that in a comment that states that the user should add `RLock()` if the method is to be accessed from another routine or when another writer gets introduced?

internal/reaper/source_reaper.go
pashagolub: We'll see :)

internal/reaper/source_reaper.go
0xgouda: Fixed in `2be6cd4`, details about the bug this might have caused can be found in the commit description.

internal/reaper/source_reaper.go
pashagolub: Sorry for delay... Still on my todo list

pashagolub: added two more and we're done. thanks!

PR #1319, opened 2026-03-20
[*] refactor `EnabledSourceSwitch`

Mazen050:
## Description

Fixes #
The enabled toggle doesn’t update after toggling enable/disable in the edit form and clicking Update.
Cause: `useState(source.IsEnabled)` only initializes once, so it gets out of sync with server updates.

The component used:
```ts
const [checked, setChecked] = useState(source.IsEnabled);
```
This initializes state only on first render, so when new server data arrives, checked does not update.

The previous useEffect workaround synced state after render but Introduced an extra render cycle

The solution is to remove local state and derive the value directly from mutation:
```ts
const checked = editEnabled.isPending ? !source.IsEnabled : source.IsEnabled;
```
this  shows optimistic toggle (`!source.IsEnabled`) while mutation is pending and after success reflects updated server value amd on error reverts automatically to original value

No extra renders are required with this aproach and no stale states. also keeps UI in sync with server.

## AI & Automation Policy

claude sonnet 4.6 helped in debuging and was used to review

## Checklist


mohamadyasser118:
Hello, You can consider making the switch disabled while the request is running (`editEnabled.isPending`) so the action doesn’t happen twice by adding `disabled={editEnabled.isPending}` to the `<Switch>`

Mazen050:
> You can consider making the switch disabled while the request is running

Sure since this is a refactor I guess it should be complete

pashagolub:
Thanks. Not needed

PR #1324, opened 2026-03-20
fix(webui): fix flex columns ignoring resize, fixes #1323

mohamadyasser118:
Fixes #1323 

I made changes: 
- `columnSizing` initializes as {}, only user-set widths are stored
- `columnsWithSizing` only overrides a column when the user has resized it, so when applying user width, strip `flex` and `minWidth` so the pixel value is respected in both directions

https://github.com/user-attachments/assets/87449b99-2f19-455c-84fa-5bdc911f5c39

## AI & Automation Policy

Drafted with the assistance Claude Sonnet 4.6

## Checklist


pashagolub:
thanks. not needed

PR #1354, opened 2026-04-02
Improve log severity parsing and extend locale support

mohamadyasser118:
Logs may not always follow strict formatting (e.g., lowercase or localized values).
I tried to ensure more consistent parsing and more accurate event counting by these changes. 

What changes:
- Normalize severity values to uppercase, so this handles cases like error, Warning correctly
- Add support for lc_messages = 'C'
- add missing French mapping (PANIQUE: PANIC)
- Improve fallback behavior for unknown values

I added more tests to cover these scenarios

## AI & Automation Policy

Drafted with Codex

## Checklist

### DIFF
--- a/internal/reaper/logparser.go
+++ b/internal/reaper/logparser.go
@@ -21,8 +21,9 @@ import (
 var pgSeverities = [...]string{"DEBUG", "INFO", "NOTICE", "WARNING", "ERROR", "LOG", "FATAL", "PANIC"}
 var pgSeveritiesLocale = map[string]map[string]string{
 	"C.": {"DEBUG": "DEBUG", "LOG": "LOG", "INFO": "INFO", "NOTICE": "NOTICE", "WARNING": "WARNING", "ERROR": "ERROR", "FATAL": "FATAL", "PANIC": "PANIC"},
+	"C":  {"DEBUG": "DEBUG", "LOG": "LOG", "INFO": "INFO", "NOTICE": "NOTICE", "WARNING": "WARNING", "ERROR": "ERROR", "FATAL": "FATAL", "PANIC": "PANIC"},
 	"de": {"DEBUG": "DEBUG", "LOG": "LOG", "INFO": "INFO", "HINWEIS": "NOTICE", "WARNUNG": "WARNING", "FEHLER": "ERROR", "FATAL": "FATAL", "PANIK": "PANIC"},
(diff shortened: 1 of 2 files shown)


0xgouda:
TIP: You should always open an issue/discussion so we can first discuss the problem/feature before doing code

From your description, I understand we have a bug in the parser. Can you provide clear and simple (and easy to read :) ) reproduction steps that would trigger this bug?

0xgouda:
Also, can you reference the postgres docs for your modifications? Is `PANIQUE` a valid word that postgres might generate? Can `ERROR, DEBUG, etc.` be lowercase?

mohamadyasser118:
> TIP: You should always open an issue/discussion so we can first discuss the problem/feature before doing code
> 
> From your description, I understand we have a bug in the parser. Can you provide clear and simple (and easy to read :) ) reproduction steps that would trigger this bug?

Thanks for your feedback, I'll make sure to open an issue first before submitting PRs in the next time. 

I added the tests the I thought it would cause an error .. 

It happens when I call `severityToEnglish("en", "error")`. The old code returns `"error"` , but downstream event counts are keyed by uppercase strings (`"ERROR"`, `"WARNING"`, etc.), so the event is silently never counted. Same applies to any locale

about the missing `"C"` localeو The old code would look up `pgSeveritiesLocale["C"]` get a nil map (not `C.`) 

To be honest I didn't test the `PANIK` locally as I don't use French locale PostgreSQL, but I just notice that the `PANIK` word is not French so I thought it was copied form the German syntax.

mohamadyasser118:
> Also, can you reference the postgres docs for your modifications? Is `PANIQUE` a valid word that postgres might generate? Can `ERROR, DEBUG, etc.` be lowercase?

I rechecked the fr.po file and didn't find `PANIQUE`, I just thougt it can prevent locale error when it is used, sorry for that 
but about `ERROR, DEBUG, etc.` I thought about the normalization as a purely defensive against log shippers or middleware that may cause errors before the logs reach pgwatch.

0xgouda:
Honestly, I don't understand what's wrong here. Can `ERROR, WARNING, etc.` be issued lowercase by postgres? can you add reproduction steps for the problem you are seeing?

Just that a specific function returns `"error"` doesn't mean anything.

mohamadyasser118:
> Honestly, I don't understand what's wrong here. Can `ERROR, WARNING, etc.` be issued lowercase by postgres? can you add reproduction steps for the problem you are seeing?
> 
> Just that a specific function returns `"error"` doesn't mean anything.

Yup you are right, I just felt that normalizing the logs this way is safer and more reassuring

I’ll close this PR for now and continue exploring and researching other areas can improve the project

PR #1366, opened 2026-04-15
[+] enhance TimescaleDB compose setup

0xgouda:
- use pg18 image.
- add persistent volume for the primary postgres service. 
- improve standby init.
- enable csv logging.


pashagolub (approved):
LGTM! Thanks

PR #1372, opened 2026-04-21
[+] add custom `TimescaleDB` Dockerfile with plpython3u support

0xgouda:
- Creates a new `Dockerfile.timescaledb-plpython3u`
- Updates `compose.timescaledb.yml` to build from it instead of using the pre-built `timescale/timescaledb` image that doesn't have plpython3u


pashagolub (approved):
LGTM!

PR #1375, opened 2026-04-22
[+] enhance `Query Performance Analysis` pg dashboard

0xgouda:
- Fix the panel queries' logic to use the `sum(col - col_lag)` pattern used in other dashboards instead of `max(col) - min(col)`, which skips counting the changes between the elements at the boundaries of each time segment.
- Fix the panel queries' logic to show the top $n for each time segment, not across the entire user-specified range.
- Address #1152 for this dashboard
- Change the panels to have a stacked view, which is helpful for deducing the cumulative load of the top $n queries
- Make panels display the rate per call/second instead of raw value increase (similar to prometheus `rate()`).

Screenshots (the updated version is hosted on the dc1 vm and is probably better to inspect it from there) :

 
image
image
image

---

TODO:
- [x] ask for review by our dbas.
- [ ] evaluate how slow the new queries might be on full 2-weeks data, for both timescaledb and vanilla postgresql setups.

### DIFF
--- a/grafana/postgres/v12/3-query-performance-analysis.json
+++ b/grafana/postgres/v12/3-query-performance-analysis.json
@@ -19,7 +19,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 42,
+  "id": 6,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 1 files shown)


pashagolub (approved):
Looks great!

Issue #1377, opened 2026-04-24
Unreachable YAML source prevents metrics collection start for all databases


rcapliez:
**Describe the bug**

**To Reproduce**

Configure pgwatch3 (v3.7.0) to use a YAML file with multiple database sources.

Include at least one source that is guaranteed to fail (e.g., a non-existent hostname or an offline instance).

Start the pgwatch3 pod/binary.

Observe the logs: the sink activates, but the fetcher never starts. The logs will show repeated connection errors and the message could not refresh active sources.

**Expected behavior**
pgwatch3 should be resilient to individual source failures. If a database is unreachable during the source refresh cycle, it should:

Log the error for that specific source.

Mark that source as "unhealthy" or skip it for the current cycle.

Proceed to start the metrics collection (fetcher) for all other reachable databases.
**Screenshots**
```
2026-04-23 15:08:52.930 [INFO] [sink:postgres] [db:metricsdb] measurements sink is activated
2026-04-23 15:08:57.955 [ERROR] [database:defaultdb] [err:failed to connect to `user=pgwatch database=defaultdb`: 1.2.3.4:12345 (db-prod-instance-01.h.aivencloud.com): dial error: timeout: context deadline exceeded]
2026-04-23 15:08:57.955 [ERROR] [err:failed to connect to `user=pgwatch database=defaultdb`: 1.2.3.4:12345 (db-prod-instance-01.h.aivencloud.com): dial error: timeout: context deadline exceeded] [time:5.01624140]
2026-04-23 15:09:26.283 [ERROR] [error:failed to connect to `user=pgwatch database=defaultdb`: 1.2.3.4:12345 (db-prod-instance-01.h.aivencloud.com): dial error: timeout: context deadline exceeded] could not refresh active sources, using last valid cache
2026-04-23 15:09:26.289 [INFO] [metrics:75] [presets:17] metrics and presets refreshed
2026-04-23 15:11:31.309 [ERROR] [database:defaultdb] [err:failed to connect to `user=pgwatch database=defaultdb`: 5.6.7.8:12345 (db-dev-instance-99.aivencloud.com): server error: FATAL: remaining connection slots are reserved for roles with the SUPERUSER attribute]
[...]

PR #1378, opened 2026-04-24
[+] implement parallel source discovery

pashagolub:
Improves dead-source handling with parallel resolution and `instance_up=0` on discovery failure.

`Sources.ResolveDatabases()` previously resolved each source sequentially. A single slow or unresponsive source (e.g. a continuous-discovery endpoint behind a firewall) would block discovery of all subsequent sources for the full connection timeout duration.

Sources are now resolved concurrently using `sync.WaitGroup.Go()`. Results are collected into a pre-allocated indexed slice to preserve deterministic ordering. Per-source error logging with source name is included in the resolver itself.

When a `SourcePostgresContinuous` or `SourcePatroni` source fails to resolve any databases, `LoadSources()` now emits `instance_up=0` to the configured sinks. This makes the failure visible in dashboards and alerting, consistent with how unreachable directly-monitored sources are handled.

### DIFF
--- a/internal/sources/resolver.go
+++ b/internal/sources/resolver.go
@@ -28,7 +28,7 @@ import (
 
 // ResolveDatabases() updates list of monitored objects from continuous monitoring sources, e.g. patroni.
 // Each source is resolved concurrently so that a slow or unreachable source does not block the others.
-func (srcs Sources) ResolveDatabases() (_ SourceConns, err error) {
+func (srcs Sources) ResolveDatabases(onError func(string)) (_ SourceConns, err error) {
 	type result struct {
 		dbs SourceConns
 		err error
@@ -45,6 +45,9 @@ func (srcs Sources) ResolveDatabases() (_ SourceConns, err error) {
 	resolvedDbs := make(SourceConns, 0, len(srcs))
 	for i, res := range results {
 		if res.err != nil {
+			if onError != nil {
+				onError(srcs[i].Name)
+			}
 			logger.WithField("source", srcs[i].Name).WithError(res.err).Error("could not resolve databases from source")
 			err = errors.Join(err, res.err)
 		}
--- a/internal/sources/resolver.go
+++ b/internal/sources/resolver.go
@@ -14,6 +14,7 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	jsoniter "github.com/json-iterator/go"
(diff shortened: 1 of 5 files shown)


0xgouda:
Is this related to #1377?

pashagolub:
> Is this related to #1377?

In some way. I couldn't reproduce that issue but I found that misconfigured discovery sources could cause a huge delays.

internal/sources/resolver.go
```diff
@@ -25,16 +26,32 @@ import (
...
+	wg.Wait()
 	resolvedDbs := make(SourceConns, 0, len(srcs))
-	for _, s := range srcs {
-		if !s.IsEnabled {
-			continue
+	for i, res := range results {
+		if res.err != nil {
+			if onError != nil {
+				onError(srcs[i].Name)
+			}
```
0xgouda, line 50: I have a concern here that can be reproduced with the following steps:

1. Define a target that happens to be unreachable and is of the kind `postgres-continuous-discovery`
2. pgwatch writes `instance_up=0` for the target for a while with `dbname = sourceName`
3. The target becomes alive
4. pgwatch runs for a while and now writes the updated `instance_up = 1`, but with a new `dbname = sourceName + _ + realDbname`
5. The target is down again and its `instance_up = 0` is written with `dbname = sourceName + _ + realDbname`

So the full instance uptime history becomes a bit disconnected, with different dbname[s].

But generally, I think that's the best we can do, just wanted to note this behaviour.

pashagolub: Good catch! We could use `source` instead of `dbname`. This way we will know for sure at which point that happened

0xgouda: Can you explain more?

pashagolub:
```go
// WriteInstanceDown writes instance_up = 0 metric to sinks for the given source
func (r *Reaper) WriteInstanceDown(md *sources.SourceConn) {
	r.measurementCh <- metrics.MeasurementEnvelope{
		DBName:     md.Name,
		MetricName: specialMetricInstanceUp,
		Data: metrics.Measurements{metrics.Measurement{
			metrics.EpochColumnName: time.Now().UnixNano(),
			"kind": string(md.Kind),
		//  ^^^^^^^^^^^^^^^^^^^^^^^
			specialMetricInstanceUp: 0},
		},
	}
}
```

pashagolub: this way grafana could distinguish regular databases vs all other

0xgouda: I don't think adding the `kind` to the measurements will help. 

What if we have 2+ `postgres-continuous-discovery`/`patroni` sources? Grouping by `kind` then will be useless.

0xgouda: And again, I guess what we have already is the best we can do, since we won't be able to add any instance identifiers (i.e., `sys_id`) to the measurements before connecting at least once.

But then, in Grafana, one might have 2 panels like:
`instance uptime history grouped by sys_id`
and
`instance uptime history for instances with unknown sys_id`

and they should complement each other, to give the full uptime history view.

PR #1379, opened 2026-04-27
[*] improve `custom_installation.md` docs file

0xgouda:
Based on internal and [external](https://github.com/cybertec-postgresql/pgwatch/discussions/1245) feedback requesting improvements for the documentation.

This rewrites the `custom_installation.md` docs file to:

1. Make it clearer and better in structure
2. Reference the Grafana docs for Grafana installation/setup steps, as Its better maintained there

### DIFF
--- a/docs/tutorial/custom_installation.md
+++ b/docs/tutorial/custom_installation.md
@@ -108,7 +108,7 @@ Create `/etc/pgwatch/sources.yaml`:
 |--------|-------------|---------|
 | `name` | Unique name for this source | `mydb` |
 | `kind` | Source type: `postgres`, `postgres-continuous-discovery`, `pgbouncer`, `pgpool`, `patroni` | `postgres` |
-| `conn_str` | PostgreSQL, etcd, Consul, or ZooKeepr connection string | `postgresql://user:pass@host/db` or `etcd://host1:1234,host2:1344/scope/member` |
+| `conn_str` | PostgreSQL or etcd connection string | `postgresql://user:pass@host/db` or `etcd://host1:1234,host2:1344/scope/member` |
 | `preset_metrics` | Preset to use: `minimal`, `basic`, `exhaustive`, `unprivileged`, etc. | `exhaustive` |
 | `custom_metrics` | Custom metrics with intervals (seconds) | `{ backends: 300 }` |
 | `include_pattern` | Regex to filter databases (for continuous discovery) | `^mydb_` |
--- a/docs/tutorial/custom_installation.md
+++ b/docs/tutorial/custom_installation.md
@@ -2,379 +2,205 @@
 title: Custom installation
 ---
 
-As described in the [Components](../concept/components.md)
-chapter, there is a couple of ways how to set up pgwatch.
-Two most common ways though are using the central **configuration database**
-approach and the **YAML file** based approach, plus Grafana to
-visualize the gathered metrics.
+This chapter describes how to set up pgwatch manually, giving you full control over each component. For a simpler setup, see the [Docker installation](docker_installation.md) guide.
 
-## Configuration Database based setup
+## Overview
 
-### Overview of installation steps
+pgwatch consists of three main components:
 
-1. Install Postgres or use any available existing instance - v14+
-    is required but the latest major version is recommended.
-1. Bootstrap the configuration database.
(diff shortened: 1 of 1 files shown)


docs/tutorial/custom_installation.md
```diff
@@ -2,379 +2,205 @@
...
-        ```
-
-1. **Bootstrap the measurements storage database (sink)**
-
-    !!! note
-        The detailed steps are described in the
-        [Bootstrapping the Metrics Measurements Database (Sink)](../howto/metrics_db_bootstrap.md) chapter
+- **Using apt (Debian/Ubuntu)**
+    ```bash
+    # Follow instructions from: https://wiki.postgresql.org/wiki/Apt#Quickstart to add the official PostgreSQL apt repoistory
```
Mazen050:
```suggestion
    # Follow instructions from: https://wiki.postgresql.org/wiki/Apt#Quickstart to add the official PostgreSQL apt repository 
```
typo :)

RafayKhattak:
Hey @0xgouda , I did a thorough scan of the new changes and caught a few things. dropping them all here in one go so you can grab what you want:

**Technical / Scripting Catches:**
1. **Broken bash command:** In the "Start pgwatch" section, the `# or --sources... \` comment is placed inside a backslash-continued command. In bash, this will comment out the `--sink` line and break the shell execution.
2. **YAML syntax:** The `....` at the bottom of the `sources.yaml` example isn't valid YAML. Might be safer to just use a standard comment like `# ... more sources ...`.
3. **SQL consistency:** Step 2 uses lowercase `create user pgwatch password '...'`, but Step 4 uses the uppercase `CREATE USER pgwatch WITH PASSWORD '...';`. Might be nice to align them to the uppercase style for consistency.
4. **Outdated Go URL:** The comment `# Install Go` points to `https://golang.org/doc/install`, which now redirects. Updating it to `https://go.dev/doc/install` avoids the redirect hop.
 
**Formatting & Copy Nits:**
5. **Overview count:** The text says "three main components" but the numbered list right below it actually has four items (Visualization is the 4th).
6. **Stray colon:** In the same list, item 2 has a stray colon (`**Configuration store**: - Where you define...`) while the others just use a dash.
7. **Bullet capitalization:** Under the "Installation Methods" bullets, some descriptions start with lowercase (`store monitored databases...`) while others start with uppercase (`Use the pgwatch web UI...`). 
8. **Hyphenation:** In the Grafana note at the end, `builtin` should probably be hyphenated as `built-in`.
 
Feel free to ignore the minor formatting nits, but definitely watch out for that bash script continuation! Hope this helps :)

docs/tutorial/custom_installation.md
```diff
@@ -2,379 +2,205 @@
...
+## Installation Methods
 
-1. **Install Postgres**
+Choose how you want to manage your monitoring configurations:
 
-    Follow the standard Postgres install procedure basically. Use the
-    latest major version available, but minimally v14+ is required.
+1. **PostgreSQL Database**  
+    - store monitored databases and metrics configs in a PostgreSQL database.  
+    - Use the pgwatch web UI or REST API to manage it.
```
pashagolub: We can use webui and REST API for all kind of configurations

docs/tutorial/custom_installation.md
```diff
@@ -2,379 +2,205 @@
...
-    some more extensions or privileges are missing on the monitored
-    database side. See the [according chapter](preparing_databases.md).
+**sources configuration options**:
 
-1. **Install Grafana**
+| Option | Description | Example |
+|--------|-------------|---------|
+| `name` | Unique name for this source | `mydb` |
+| `kind` | Database type: `postgres`, `postgres-continuous-discovery`, `pgbouncer`, `pgpool`, `patroni` | `postgres` |
+| `conn_str` | PostgreSQL connection string | `postgresql://user:pass@host/db` |
```
pashagolub: or etcd connection string, e.g. `etcd://host1:1234,host2:1344/scope/member

docs/tutorial/custom_installation.md
0xgouda: I know, I just thought they would usually be used with configs that use a pg db.
But ok, will fix.

pashagolub: probably they are but I bet some users will roast us for that :)

pashagolub (approved):
Thanks for working on this!

PR #1382, opened 2026-04-29
[-] apply non-recursive yaml dir reading, fixes #1368

### LINKED ISSUE #1368: Directory-based YAML source loading reads hidden Kubernetes mount internals and causes duplicate sources
### Problem definition
Currently, there are 3 ways to provide sources to pgwatch: Postgres URI, a YAML file or a folder of YAML files containing info on which DBs to monitor.

When PGwatch is configured with a directory of YAML source files, it recursively [walks](https://github.com/cybertec-postgresql/pgwatch/blob/master/internal/sources/yaml.go#L99) the directory and loads all YAML files.

I'm deploying pgwatch on Kubernetes using [pgwatch helm chart](https://github.com/cybertec-postgresql/pgwatch-charts/tree/686c0768ae517b0253656872a1fb263f56f8e4ab/helm/pgwatch). I've created a ConfigMap from the user provided YAML files and mounted them inside a directory in pgwatch pod. 

The created files inside the pod are:
```
/tmp/pgwatch-sources $ ls
custom-sources.yaml
/tmp/pgwatch-sources $ ls -al
total 12
drwxrwsrwx    3 root     1000          4096 Apr 15 19:35 .
drwxrwxrwt    1 root     root          4096 Apr 15 19:35 ..
drwxr-sr-x    2 root     1000          4096 Apr 15 19:35 ..2026_04_15_19_35_17.1518305580
lrwxrwxrwx    1 root     1000            32 Apr 15 19:35 ..data -> ..2026_04_15_19_35_17.1518305580
lrwxrwxrwx    1 root     1000            26 Apr 15 19:35 custom-sources.yaml -> ..data/custom-sources.yaml
```

In Kubernetes, when a ConfigMap is mounted as a directory, hidden internal directories (e.g. ..data and timestamped directories created by the atomic writer) are present inside the mount. PGwatch traverses these hidden directories as well, causing the same YAML files to be read multiple times. The whole atomic writing algorithm is described [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/atomic_writer.go#L86-L138).

This results in `duplicate source with name '%s' found` during [validate](https://github.com/cybertec-postgresql/pgwatch/blob/master/internal/sources/types.go#L65) function call.

To reproduce the issue, you can manually create a configmap and mount it to the pgwatch pod running on Kubernetes. After that, provide this directory as source to pgwatch and check pgwatch logs. 
Alternatively, you can deploy the helm-chart mentioned above to your Kubernetes cluster with 
```
helm install pgwatch -n pgwatch --create-namespace \ 
  -f values.yaml \
  --set-file "pgwatch.sources.files.custom-sources\.yaml=custom-sources.yaml" \
  .
```
and observe. 

### Possible Solutions

From the user perspective, the YAML files are provided once. Hidden/internal directories created by the underlying filesystem or orchestration problem should not result in duplicate source loading.
From the developer perspective, this is an implementation detail of how Kubernetes mounts ConfigMaps. That's not necessarily a problem that pgwatch should provide a solution. 

I'd ask for finding a middle ground: There would be several possible solutions:

1. Ignoring hidden directories(dirs starting with `.`) during directory traversal.  
2. Provide a configuration flag to disable traversal of hidden directories. (Users can decide)  
3. (Less preffered) Ignore the `..data` symlink and `..timestamp` dir during traversal. This one is very Kubernetes specific, probably doesn't fit to pgwatch's nature. 

I'd propose 1st or the 2nd option, rather a generic solution than a Kubernetes specific one.

pashagolub:
replace `WalkDir` with `filepath.Glob`, those functions never intended to be recursive in the first place

### DIFF
--- a/internal/metrics/yaml.go
+++ b/internal/metrics/yaml.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"gopkg.in/yaml.v3"
@@ -64,13 +65,16 @@ func (fmr *fileMetricReader) getMetrics() (metrics *Metrics, err error) {
 	}
 	switch mode := fi.Mode(); {
 	case mode.IsDir():
-		var matches []string
-		if matches, err = filepath.Glob(filepath.Join(fmr.path, "*.y*ml")); err != nil {
+		var entries []os.DirEntry
+		if entries, err = os.ReadDir(fmr.path); err != nil {
 			return nil, err
 		}
-		for _, path := range matches {
+		for _, entry := range entries {
+			if ext := strings.ToLower(filepath.Ext(entry.Name())); ext != ".yaml" && ext != ".yml" {
+				continue
+			}
 			var m *Metrics
-			if m, err = fmr.loadMetricsFromFile(path); err != nil {
+			if m, err = fmr.loadMetricsFromFile(filepath.Join(fmr.path, entry.Name())); err != nil {
 				return nil, err
 			}
 			maps.Copy(metrics.PresetDefs, m.PresetDefs)
--- a/internal/metrics/yaml.go
+++ b/internal/metrics/yaml.go
@@ -3,11 +3,9 @@ package metrics
 import (
 	"context"
 	_ "embed"
-	"io/fs"
 	"maps"
 	"os"
 	"path/filepath"
-	"strings"
 	"sync"
 
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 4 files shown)


internal/metrics/yaml.go
```diff
@@ -66,21 +64,18 @@ func (fmr *fileMetricReader) getMetrics() (metrics *Metrics, err error) {
...
-		err = filepath.WalkDir(fmr.path, func(path string, d fs.DirEntry, err error) error {
-			if err != nil {
-				return err
-			}
-			ext := strings.ToLower(filepath.Ext(d.Name()))
-			if d.IsDir() || ext != ".yaml" && ext != ".yml" {
-				return nil
-			}
+		var matches []string
+		if matches, err = filepath.Glob(filepath.Join(fmr.path, "*.y*ml")); err != nil {
```
serdardalgic: Hello! 👋 
Maybe I'm being a bit picky here, but I just wanted to point out this `*.y*ml` can match more than `.yml` and `.yaml`, which could match unintended files. 

Would it make sense to check the extension explicitly instead? A helper based on strings.ToLower(filepath.Ext(name)) would be more explicit and could also be reused in both changed files.

This also seems to be the pattern other Go programs/libraries use. To name a few: 

* prometheus [promtool](https://github.com/prometheus/prometheus/blob/main/cmd/promtool/main.go#L829-L837)
* helm [v3](https://github.com/helm/helm/blob/main/internal/chart/v3/chart.go#L176-L179)
* tfsec [trivy](https://github.com/aquasecurity/tfsec/blob/master/internal/pkg/config/config.go#L37-L47)

internal/sources/yaml.go
```diff
@@ -103,20 +103,17 @@ func (fcr *fileSourcesReaderWriter) getSources() (dbs Sources, err error) {
...
-		err = filepath.WalkDir(fcr.path, func(path string, d fs.DirEntry, err error) error {
-			if err != nil {
-				return err
-			}
-			ext := strings.ToLower(filepath.Ext(d.Name()))
-			if d.IsDir() || ext != ".yaml" && ext != ".yml" {
-				return nil
-			}
+		var matches []string
+		if matches, err = filepath.Glob(filepath.Join(fcr.path, "*.y*ml")); err != nil {
```
serdardalgic: Same here https://github.com/cybertec-postgresql/pgwatch/pull/1382/changes#r3167681018

internal/metrics/yaml.go
pashagolub: yes, I know. I can live with that. Glob is accepting only pattern :) I just can say that this pattern is used so name your files properly! Thanks anyway for your attention to details

internal/sources/yaml.go
pashagolub: from fucntionality perspective, does it solve your issue?

internal/metrics/yaml.go
serdardalgic: Got it, makes sense :+1: thanks for clarifying! 
The non-recursive change should probably solve the issue without any problems. Testing it now.

internal/sources/yaml.go
serdardalgic: oh yeah, I think so. I'm testing it now and will share the results with you.

internal/metrics/yaml.go
pashagolub: You really got through to me! :) Fixed it!

PR #1405, opened 2026-05-11
[!] add Prometheus exporter as a pgwatch source

pashagolub:
Allow pgwatch to scrape arbitrary Prometheus exporters, apply family-level filtering, and write measurements to any configured sink — including acting as a low-overhead proxy when the sink is a Prometheus endpoint.


0xgouda:
Won't this be an unnecessary additional complexity, users can just use pgwatch as a prometheus exporter - which is a logical thing to do to avoid unnecessarily managing additional solution[s] - ?.

pashagolub:
> Won't this be an unnecessary additional complexity, users can just use pgwatch as a prometheus exporter - which is a logical thing to do to avoid unnecessarily managing additional solution[s] - ?.

I believe you miss the whole point of this feature. This is to fetch measurements from Prometheus, e.g. [patroni](https://patroni.readthedocs.io/en/latest/rest_api.html#monitoring-endpoint)

docs/howto/monitor_prometheus_exporter.md
pashagolub, line 1: @0xgouda check this out! :)

PR #1408, opened 2026-05-14
[*] migrate to time-only partitioning in pg sink

### LINKED ISSUE #1392: Support metric-time only partitioning for postgres sink
As discussed in https://github.com/cybertec-postgresql/pgwatch/discussions/1365#discussioncomment-16814482, metric-dbname-time partitioning in setups with a large number of monitored dbs and a single pg metrics database results in a huge number of partitions, which heavily affects performance due to planning overhead, and it also affects backup and upgrade operations badly.

### LINKED ISSUE #1409: Switch to time-only partitioning

0xgouda:
- Simplified the partitioning schema by transitioning from 3-level (metric-dbname-time) to 2-level (metric-time) partitioning
    - This solves the performance issues caused by having thousands of partitions for setups with extremely large numbers of monitored databases
- Removed `drop_source_partitions` function
- The schema migration will 
    1. rename the metric table to `*_before_v6_migration`
    2. create a metric table with the original name and time-only partitioning
    3. loop on all partitions for every renamed metric table and:
        1. insert the data into the new table 
        2. detach the partition 
        3. drop it.
    - partitions in the new table will have a `1 day` partitioning interval for the old data.
- Change `--partition-interval` 's default to `1 day`
    
 Closes: #1409 
 Closes: #1392
 
 ---
 
 - Benchmarked the migration time on a sink database with 100 monitored sources, all built-in metrics used (i.e., 80), around 200_000 records, and 501 partitions in each metric table, and it took **1:16:13.67** to run.
     - Bigger setups with thousands of monitored dbs will take much longer.

### DIFF
--- a/internal/sinks/postgres.go
+++ b/internal/sinks/postgres.go
@@ -61,16 +61,16 @@ var (
 // However, one is able to use any Postgres-compatible database as a storage backend,
 // e.g. PGEE, Citus, Greenplum, CockroachDB, etc.
 type PostgresWriter struct {
-	ctx                      context.Context
-	sinkDb                   db.PgxPoolIface
-	metricSchema             DbStorageSchemaType
-	opts                     *CmdOpts
-	retentionInterval        time.Duration
-	maintenanceInterval      time.Duration
-	input                    chan metrics.MeasurementEnvelope
-	lastError                chan error
-	forceRecreatePartitions  bool                                        // to signal override PG metrics storage cache
-	partitionMapMetric 		 map[string]ExistingPartitionInfo // metric = min/max bounds
+	ctx                     context.Context
+	sinkDb                  db.PgxPoolIface
+	metricSchema            DbStorageSchemaType
+	opts                    *CmdOpts
+	retentionInterval       time.Duration
+	maintenanceInterval     time.Duration
+	input                   chan metrics.MeasurementEnvelope
+	lastError               chan error
+	forceRecreatePartitions bool                             // to signal override PG metrics storage cache
+	partitionMapMetric      map[string]ExistingPartitionInfo // metric = min/max bounds
 }
 
 // make sure *dbMetricReaderWriter implements the Migrator interface
@@ -90,13 +90,13 @@ func NewWriterFromPostgresConn(ctx context.Context, conn db.PgxPoolIface, opts *
 	l := log.GetLogger(ctx).WithField("sink", "postgres").WithField("db", conn.Config().ConnConfig.Database)
 	ctx = log.WithLogger(ctx, l)
 	pgw = &PostgresWriter{
-		ctx:                      ctx,
-		opts:                     opts,
-		input:                    make(chan metrics.MeasurementEnvelope, cacheLimit),
-		lastError:                make(chan error),
-		sinkDb:                   conn,
-		forceRecreatePartitions:  false,
-		partitionMapMetric: 	  make(map[string]ExistingPartitionInfo),
+		ctx:                     ctx,
+		opts:                    opts,
+		input:                   make(chan metrics.MeasurementEnvelope, cacheLimit),
+		lastError:               make(chan error),
+		sinkDb:                  conn,
+		forceRecreatePartitions: false,
+		partitionMapMetric:      make(map[string]ExistingPartitionInfo),
 	}
 	l.Info("initialising measurements database...")
 	if err = pgw.init(); err != nil {
@@ -453,7 +453,7 @@ func (pgw *PostgresWriter) EnsureMetricTimescale(pgPartBounds map[string]Existin
 	return
 }
 
-func (pgw *PostgresWriter) EnsureMetricTimePartsExist(metricPartBounds map[string]ExistingPartitionInfo) (error) {
+func (pgw *PostgresWriter) EnsureMetricTimePartsExist(metricPartBounds map[string]ExistingPartitionInfo) error {
 	var err error
 	var rows pgx.Rows
 	sqlEnsure := `select * from admin.ensure_partition_metric_time($1, $2, $3)`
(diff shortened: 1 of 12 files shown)


internal/sinks/sql/admin_schema.sql
```diff
@@ -95,5 +95,6 @@ INSERT INTO
     admin.migration (id, version)
 VALUES
     (0,  '01110 Apply postgres sink schema migrations'),
-    (1,  '01180 Apply admin functions migrations for v5');
+    (1,  '01180 Apply admin functions migrations for v5'),
+    (2,  '01409 Apply postgres sink partitioning scheme migrations');
```
pashagolub: it should have clearer name, e.g. "switch to time partitioning". Just applying scheme migrations is the process we do for every migration :)

internal/sinks/sql/ensure_partition_postgres.sql
```diff
@@ -46,6 +47,12 @@ BEGIN
 
   PERFORM pg_advisory_xact_lock(regexp_replace( md5(metric) , E'\\D', '', 'g')::varchar(10)::int8);
 
+  -- Check if table exists and has LIST partitioning (legacy); if so, drop it for conversion to RANGE
+  SELECT to_regclass('public.' || quote_ident(metric)) INTO l_table_oid;
+  IF EXISTS (SELECT 1 FROM pg_partitioned_table WHERE partrelid = l_table_oid AND partstrat = 'l')
+     AND l_table_oid IS NOT NULL THEN
```
pashagolub:
```sql
IF EXISTS (SELECT 1 FROM pg_partitioned_table WHERE partrelid = to_regclass('public.' || quote_ident(metric)) AND partstrat = 'l') THEN
```
is enough. NULL never satisfies `=` operator

internal/sinks/postgres.go
```diff
@@ -593,6 +593,21 @@ var migrations func() migrator.Option = func() migrator.Option {
 			},
 		},
 
+		&migrator.Migration{
+			Name: "01409 Apply postgres sink partitioning scheme migrations",
```
pashagolub: Same here. Name is too general

0xgouda:
@pashagolub 

1. What are your thoughts on the drop-on-write trick I used? Do you think it may cause any problems?

2. Also, I am thinking of changing the defaults for `--partition-interval` and `--retention` to say `3 days`/`15 days` to make the time partitions pruning more effective. What do you think?
    - As if we sticked to the current `7 days`/`14 days` defaults, we will only have 2 partitions at a time, which I believe isn't that effective.

pashagolub:
What is drop on write?
I'd go for 1d partition and 2w retention

internal/sinks/sql/ensure_partition_postgres.sql
```diff
@@ -50,31 +46,28 @@ BEGIN
...
+      (
+        SELECT 1 
+        FROM pg_partitioned_table 
+        WHERE 
+          partrelid = to_regclass('public.' || quote_ident(metric))
+          AND
+          partstrat = 'l'
+      ) THEN
+    EXECUTE format('DROP TABLE public.%I CASCADE', metric);
+  END IF;
```
0xgouda: @pashagolub this.

0xgouda: Then the rest of the script should run normally, creating the table, its partitions, etc.

pashagolub: but if we drop partition we lose the data, right?

0xgouda: Yeah, we will drop the entire table, but only as part of the migration to v6.

Should we have a meeting to discuss it?

internal/sinks/sql/ensure_partition_postgres.sql
```diff
@@ -50,31 +46,15 @@ BEGIN
...
-
   -- 1. level
   IF to_regclass('public.' || quote_ident(metric)) IS NULL
   THEN
-    EXECUTE format('CREATE TABLE public.%I (LIKE admin.metrics_template INCLUDING INDEXES) PARTITION BY LIST (dbname)', metric);
+    EXECUTE format('CREATE TABLE public.%I (LIKE admin.metrics_template INCLUDING INDEXES) PARTITION BY RANGE (time)', metric);
     EXECUTE format('COMMENT ON TABLE public.%I IS $$pgwatch-generated-metric-lvl$$', metric);
   END IF;
 
   -- 2. level
```
pashagolub, line 56: should we remove this comment?

internal/sinks/postgres.go
```diff
@@ -631,6 +593,31 @@ var migrations func() migrator.Option = func() migrator.Option {
 			},
 		},
 
+		&migrator.Migration{
+			Name: "01409 Switch to time-only partitioning",
+			Func: func(ctx context.Context, tx pgx.Tx) error {
+				_, err := tx.Exec(ctx, `SELECT admin.drop_all_metric_tables()`)
```
pashagolub: Thinking out loud.. What if we make a copy of current measurements with `ALTER SCHEMA public RENAME TO backup`, then create all new tables with new partitioning and then `INSERT INTO public.<new-table> SELECT * FROM backup.<old-table>`? Then we keep the data and users are happy

internal/sinks/postgres.go
```diff
@@ -637,6 +593,111 @@ var migrations func() migrator.Option = func() migrator.Option {
...
+					suffix := "_before_v6_migration"
+					if strings.HasSuffix(metricTable, suffix) {
+						continue
+					}
+					metricTableRenamed := metricTable + suffix
+
+					err = pgx.BeginFunc(ctx, conn, func(tx pgx.Tx) error {
+						// check if the table is already migrated to avoid errors on re-run after a failed migration attempt
+						var isTableMigrated bool
+						if err = conn.QueryRow(ctx, `SELECT true pg_partitioned_table WHERE partrelid = to_regclass($1) AND partstrat = 'r'`, metricTable).Scan(&isTableMigrated); isTableMigrated || err != nil {
```
pashagolub: `SELECT true FROM pg_partitioned_table...`?

pashagolub: You are using `BeginFunc` but ignores `tx` which should be used inside the func. Either you shouldn't use `BeginFunc` or use `tx` instead of `conn`

pashagolub: if we cannot find the table the result set will be empty and err will have `pgx.ErrNoRows`, function exits, migration failed

pashagolub: We could use `EXISTS` probably here

internal/sinks/postgres.go
```suggestion
const MigrationsCount = 3
```
we'll have 3 migrations

0xgouda: Why do we have this variable in the first place? I don't see it used anywhere.

pashagolub (approved):
LGTM

PR #1415, opened 2026-05-20
[-] fix deadlock in `SourceReaper`

### LINKED ISSUE #1412: deadlock in `SourceReaper`
**Describe the bug**
A deadlock that happens in the new `SourceReaper` module which blocks the collection of metrics for that source.

**To Reproduce**
1. Run pgwatch built from master with `--max-parallel-connections-per-db=1`
2. Modify one of your configured metrics to have invalid sql, so that it would trigger a retry for itself + at least 1 more metric. 
3. Observe the deadlock

**Expected behavior**
There shouldn't be any deadlocks that block the source monitoring.

**Screenshots**
- Note: There is no single `measurements fetched` log message generated for that source again at all.
```
2026-05-18 20:45:00.944 [INFO] [sink:postgres] [db:postgres] initialising measurements database...
2026-05-18 20:45:00.951 [INFO] [sink:postgres] [db:postgres] measurements sink is activated
2026-05-18 20:45:00.952 [INFO] [sources:1] sources refreshed
2026-05-18 20:45:00.956 [INFO] [metrics:74] [presets:15] metrics and presets refreshed
2026-05-18 20:45:00.964 [INFO] [source:test] [recovery:false] Connect OK. Version: PostgreSQL 16.13 (Ubuntu 16.13-1.pgdg24.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0, 64-bit
2026-05-18 20:45:00.968 [INFO] [source:test] starting source reaper
2026-05-18 20:45:01.025 [INFO] [source:test] [metric:instance_up] [rows:1] measurements fetched
2026-05-18 20:45:01.057 [INFO] [source:test] [metric:table_stats] [rows:151] measurements fetched
2026-05-18 20:45:01.057 [INFO] [source:test] [metric:db_size] [rows:1] measurements fetched
2026-05-18 20:45:01.057 [INFO] [source:test] [metric:bgwriter] [rows:1] measurements fetched
2026-05-18 20:45:01.057 [INFO] [source:test] [metric:postgres_role] [rows:1] measurements fetched
2026-05-18 20:45:01.082 [INFO] [source:test] [metric:stat_statements] [rows:26] measurements fetched
2026-05-18 20:45:01.082 [INFO] [source:test] [metric:locks_mode] [rows:8] measurements fetched
2026-05-18 20:45:01.082 [ERROR] [source:test] [sql:; -- covered by bgwriter] [args:[]] [err:pipeline: QueryStatement: received unexpected message type *pgproto3.EmptyQueryResponse] [pid:139671] BatchQuery
2026-05-18 20:45:01.929 [INFO] [sink:postgres] [db:postgres] [rows:189] [elapsed:16.437552ms] measurements written
```
**Additional context**
This happens because the batch closing in `executeBatch()` is defered here:
https://github.com/cybertec-postgresql/pgwatch/blob/64be56c3b3c4f23a42d37a1bceb404b397cd9abe/internal/reaper/source_reaper.go#L245-L245

So subsequent retries will block trying to acquire the connection that will never get released:
https://github.com/cybertec-postgresql/pgwatch/blob/64be56c3b3c4f23a42d37a1bceb404b397cd9abe/internal/reaper/source_reaper.go#L261-L262

0xgouda:
- Ensure the batch releases the connection before any individual retries.
- Use a hardcoded `MaxConns` of 1 for the conn pool and deprecate `--max-parallel-connections-per-db`

Closes: #1412

### DIFF
--- a/docs/reference/cli_env.md
+++ b/docs/reference/cli_env.md
@@ -39,11 +39,6 @@ It reads the configuration from the specified sources and metrics, then begins c
     Smaller size databases will be ignored and not monitored until they reach the threshold (default: 0).  
     ENV: `$PW_MIN_DB_SIZE_MB`
 
-- `--max-parallel-connections-per-db=`
-
-    Max parallel metric fetches per monitored database. Note the multiplication effect on multi-DB instances (default: 4).  
-    ENV: `$PW_MAX_PARALLEL_CONNECTIONS_PER_DB`
-
 - `--try-create-listed-exts-if-missing=`
 
     Try creating the listed extensions (comma sep.) on first connect for all monitored DBs when missing. Main usage - pg_stat_statements.  
--- a/internal/cmdopts/cmdconfig_integration_test.go
+++ b/internal/cmdopts/cmdconfig_integration_test.go
@@ -47,7 +47,7 @@ func TestConfigUpgrade_VerifyNoCircularDependency(t *testing.T) {
 	// After successful upgrade, InitConfigReaders should succeed
 	opts2 := &Options{
 		Metrics: metrics.CmdOpts{Metrics: connStr},
-		Sources: sources.CmdOpts{Sources: connStr, Refresh: 120, MaxParallelConnectionsPerDb: 1},
+		Sources: sources.CmdOpts{Sources: connStr, Refresh: 120},
 		Sinks:   sinks.CmdOpts{Sinks: []string{connStr}},
 	}
 	err = opts2.InitConfigReaders(ctx)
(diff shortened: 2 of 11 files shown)


pashagolub:
> * Use a hardcoded `MaxConns` of 1 for the conn pool and deprecate `--max-parallel-connections-per-db`

This is strong NO from me. Why would we want to set `MaxConns` to 1? We do some internal calls that may benefit from several connections

pashagolub (commented):
-1

0xgouda:
> Why would we want to set MaxConns to 1? We do some internal calls that may benefit from several connections

Yeah, you are correct, it's better to keep things as is, and if anything bad happens, it's user-controllable via `--max-parallel-connections-per-db`.

I was just trying to enforce a more lightweight behaviour on the monitored sources (i.e. using less connections).

PR #1439, opened 2026-06-16
[*] change `backends` values from int to float

avandras:
## Description

Some values of the `backends` metric, like the longest query duration, were casted to an integer, which is not very useful if you only have sub-second queries...  

Fixes #

## AI & Automation Policy

## Checklist


pashagolub (approved):
Great work!

PR #1528, opened 2026-08-18
[*] handle counter resets in dashboard queries

0xgouda:
Add Prometheus-style counter reset handling to the PostgreSQL Grafana queries.


pashagolub (approved):
LGTM

PR #1531, opened 2026-08-19
[-] fix partition creation after long downtime, closes #1529

### LINKED ISSUE #1529: Data loss due to partitioning bug
**Describe the bug**
When using a postgresql sink and on lack of partitions, pgwatch will create 4 new ones starting from the old upper bound of the last existing partitions, this is buggy and causes error if pgwatch gets down for a while - i.e. a while longer than the interval the 4 partitions will cover - and is then re-used.

This bug also exists in older releases but with the new `1 day` default partitioning interval its way more likely to happen. 

**To Reproduce**
1. Run pgwatch built from master with pg sink and let it do the init
2. Stop pgwatch and wait for a week
3. Re-run pgwatch
4. Notice the partitioning and data insertion errors that arise because pgwatch has created 3 new partitions for the days that we were off on them which don't fit the current metric dates i.e. we still don't have partitions for the collected measurements.

**Expected behavior**
Partition creation should start from the min timestamp for the inserted metrics and ignore gaps between it and the existing upper bound partition. - This will also result in having a fewer number of partitions -. 

**Screenshots**
My old pgwatch sink db partitions
```
Partitions: subpartitions.db_size_20260205 FOR VALUES FROM ('2026-02-05 00:00:00+02') TO ('2026-02-06 00:00:00+02'),
            subpartitions.db_size_20260206 FOR VALUES FROM ('2026-02-06 00:00:00+02') TO ('2026-02-07 00:00:00+02'),
            subpartitions.db_size_20260207 FOR VALUES FROM ('2026-02-07 00:00:00+02') TO ('2026-02-08 00:00:00+02'),
            subpartitions.db_size_20260208 FOR VALUES FROM ('2026-02-08 00:00:00+02') TO ('2026-02-09 00:00:00+02'),
            subpartitions.db_size_20260209 FOR VALUES FROM ('2026-02-09 00:00:00+02') TO ('2026-02-10 00:00:00+02'),
            subpartitions.db_size_20260210 FOR VALUES FROM ('2026-02-10 00:00:00+02') TO ('2026-02-11 00:00:00+02'),
            subpartitions.db_size_20260211 FOR VALUES FROM ('2026-02-11 00:00:00+02') TO ('2026-02-12 00:00:00+02')
```

When running pgwatch I keep getting: 

```
[ERROR] [sink:postgres] [db:pgwatch_metrics] [sql:select * from admin.ensure_partition_metric_time($1, $2, $3)] [args:[db_size 2026-08-19 00:49:18.295765 +0300 EEST 1 day]] [err:can't scan into dest[0] (col: part_available_from): cannot scan NULL into *time.Time][pid:84876] [time:11.678726ms] Query

2026-08-19 00:49:19.231 [WARNING] [sink:postgres] [db:pgwatch_metrics] Some metric partitions might have been removed, halting all metric storage. Trying to re-create all needed partitions on next run
2026-08-19 00:49:19.231 [ERROR] [columnNames:[time dbname data tag_data]] [err:ERROR: no partition of relation "db_size" found for row (SQLSTATE 23514)] [pid:84876] [tableName:[db_size]] [time:241.048µs] CopyFrom
2026-08-19 00:49:19.231 [ERROR] [sink:postgres] [db:pgwatch_metrics] ERROR: no partition of relation "db_size" found for row (SQLSTATE 23514)
```

And 4 new partitions get created for `1[2,5]-02-2026`

0xgouda:
When incoming data exceeds the last existing partition's upper bound, partitions will be created starting from the metric timestamp instead of the previous upper bound, avoiding NULL available ranges and missing-partition insert errors after pgwatch has been down for longer than the pre-created window.

Closes #1529

### DIFF
--- a/internal/sinks/sql/ensure_partition_postgres.sql
+++ b/internal/sinks/sql/ensure_partition_postgres.sql
@@ -10,7 +10,7 @@
 CREATE OR REPLACE FUNCTION admin.ensure_partition_metric_time(
     metric text,
     metric_timestamp timestamptz,
-    partition_period interval default '1 week'::interval,
+    partition_period interval default '1 day'::interval,
     partitions_to_precreate int default 3,
     OUT part_available_from timestamptz,
     OUT part_available_to timestamptz)
@@ -46,14 +46,11 @@ BEGIN
 
   PERFORM pg_advisory_xact_lock(regexp_replace( md5(metric) , E'\\D', '', 'g')::varchar(10)::int8);
 
-  -- 1. level
   IF to_regclass('public.' || quote_ident(metric)) IS NULL
   THEN
     EXECUTE format('CREATE TABLE public.%I (LIKE admin.metrics_template INCLUDING INDEXES) PARTITION BY RANGE (time)', metric);
     EXECUTE format('COMMENT ON TABLE public.%I IS $$pgwatch-generated-metric-lvl$$', metric);
   END IF;
-
-  -- 2. level
 
   -- Get existing partition upper bound
   SELECT max(substring(pg_catalog.pg_get_expr(c.relpartbound, c.oid, true) from 'TO \(''([^'']+)''')::timestamptz)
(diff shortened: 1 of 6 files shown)


pashagolub:
So, for example, if we have some corrupted measurement with ts far in the future, the code will create one partition starting from the last upper bound and until the measurement ts?

0xgouda:
> So, for example, if we have some corrupted measurement with ts far in the future, the code will create one partition starting from the last upper bound and until the measurement ts?

It will create only 1 partition to hold that ts; the gap between that ts partition and the last upper bound will still be empty with no partitions covering it.

internal/sinks/sql/ensure_partition_postgres.sql
pashagolub: unused var, could be removed. Not your fault though.

pashagolub (approved):
Great job! LGTM!

internal/sinks/sql/ensure_partition_postgres.sql
pashagolub:
```
ideal_length int;
```
I mean

PR #1537, opened 2026-08-28
[*] optimize `flush()` in pg sink

0xgouda:
We don't need to loop over all the rows in `Measurements` and read `epoch_ns` from each, because they typically all contain the same value extracted from `now()` in the same txn.

### DIFF
--- a/internal/metrics/types.go
+++ b/internal/metrics/types.go
@@ -134,8 +134,9 @@ func (m Measurement) GetEpoch() int64 {
 			return epoch
 		}
 	}
-	m[EpochColumnName] = time.Now().UnixNano()
-	return m[EpochColumnName].(int64)
+	t := time.Now().UnixNano()
+	m[EpochColumnName] = t
+	return t
 }
 
 type Measurements []map[string]any
--- a/internal/metrics/types.go
+++ b/internal/metrics/types.go
@@ -134,7 +134,8 @@ func (m Measurement) GetEpoch() int64 {
 			return epoch
 		}
 	}
-	return time.Now().UnixNano()
+	m[EpochColumnName] = time.Now().UnixNano()
+	return m[EpochColumnName].(int64)
 }
 
 type Measurements []map[string]any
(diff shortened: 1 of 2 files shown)


pashagolub:
cool! do you have numbers on speed imrovements?

internal/metrics/types.go
```diff
@@ -134,7 +134,8 @@ func (m Measurement) GetEpoch() int64 {
 			return epoch
 		}
 	}
-	return time.Now().UnixNano()
+	m[EpochColumnName] = time.Now().UnixNano()
+	return m[EpochColumnName].(int64)
```
pashagolub:
```suggestion
	t := time.Now().UnixNano()
	m[EpochColumnName] = t
	return t
```
is this better? w/o accessing map and type casting

0xgouda: It doesn't matter that much; this code path is never executed in normal conditions, but ok lets add it.

0xgouda:
> cool! do you have numbers on speed imrovements?

Nope, I didn't benchmark it, but I would expect an improvement only for large setups with 100s of monitored sources, with each of them returning a lot of rows.

IMO, smaller setups won't be affected that much.

pashagolub (approved):
LGTM! Good job!

0xgouda:
> cool! do you have numbers on speed imrovements?

I did a small benchmark on my local machine mimicking 1000 sources, each with 80 active metrics with the same interval, and each metric returns 200 rows, and it was 7% faster:

```
goos: linux
goarch: amd64
pkg: github.com/cybertec-postgresql/pgwatch/v6/internal/sinks
cpu: Intel(R) Core(TM) i5-10500H CPU @ 2.50GHz
               │  old.txt   │              new.txt              │
               │   sec/op   │   sec/op    vs base               │
PgSinkflush-12   131.0 ± 7%   121.9 ± 4%  -6.92% (p=0.000 n=10)
```

PR #1538, opened 2026-08-28
[-] fix the default dashboard file name

0xgouda:
The demo image and the default Grafana Compose file expect the filename to be `1-global-db-overview.json`, not `1-global-database-overview.json`; otherwise, Grafana will fail to load the default dashboard.


pashagolub (approved):
Nice catch! Approved!

PR #1541, opened 2026-08-31
[-] fix multiple logging errors

0xgouda:
1. `continue` if the metric was added to a batch Otherwise, we might print an incorrect error message
from the previous iteration if it had an `err != nil` and didn't clear it.

2. Unwrap batch execution errors to print them one at a time instead of printing a single big line for all errors.

3. Ensure all errors for non-postgres sources are collected - previously only the last error persisted and got logged -


pashagolub (approved):
LGTM

PR #1549, opened 2026-09-07
[*] optimize prom sink memory usage

0xgouda:
Don't repeat the fixed-output `strings.Join()` in the `fields` loop

A memory profile was collected before and after the change on a pgwatch test instance that ran for ~1.5 hours with 2 sources using the `exhaustive_no_python` preset and both Postgres and Prometheus sinks.

Before:
- `sinks.(*PrometheusWriter).WritePromMetrics()` consumed 85% of `alloc_space` and 90% of `alloc_objects`
- The `strings.Join()` call in `WritePromMetrics()` consumed 14% of `alloc_space`
- `alloc_space` = 18GB, `alloc_objects` = 300 millions

image

image

After:
 - `sinks.(*PrometheusWriter).WritePromMetrics()` consumed 79% of `alloc_space` and 89% of `alloc_objects`
 -  The `strings.Join()` call no longer appears in the memory profile
 - `alloc_space` = 12Gb, `alloc_objects` = 216 millions
 
 
image

### DIFF
--- a/internal/sinks/prometheus.go
+++ b/internal/sinks/prometheus.go
@@ -318,6 +318,7 @@ func (promw *PrometheusWriter) WritePromMetrics(msg metrics.MeasurementEnvelope,
 		for i, k := range labelKeys {
 			labelValues[i] = labels[k]
 		}
+		joinedLabelValues := strings.Join(labelValues, "_")
 
 		for field, value := range fields {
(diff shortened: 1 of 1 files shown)


pashagolub (approved):
Good job!

PR #1550, opened 2026-09-07
[+] cut allocations in `PrometheusWriter.WritePromMetrics`

pashagolub:
Everything that does not depend on the row is now computed once per envelope instead of once per row x field:

- fqName and the gauge/counter decision are memoized per field, and the namespace prefix is concatenated once instead of via fmt.Sprintf;
- prometheus.Desc is cached per (fqName, label keys). NewDesc allocates ~6+2N objects and runs regexp validation and two sorts on every call;
- numeric columns are converted with a direct float64(t) conversion instead of strconv.ParseFloat(fmt.Sprintf("%v", v)); only float32 values may differ at the ULP level;
- the dead labels map allocation and the map[bool]float64 literal are gone.

BenchmarkWritePromMetrics (500 rows, 3 tags, 8 columns):
  before  10.5 ms/op   5.14 MB/op   141k allocs/op
  after    6.3 ms/op   3.70 MB/op    86k allocs/op

### DIFF
--- a/internal/sinks/prometheus.go
+++ b/internal/sinks/prometheus.go
@@ -7,7 +7,6 @@ import (
 	"net"
 	"net/http"
 	"slices"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -16,6 +15,7 @@ import (
 	"github.com/cybertec-postgresql/pgwatch/v6/internal/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/model"
 )
 
 type PromMetricCache = map[string]map[string]metrics.MeasurementEnvelope // [dbUnique][metric]lastly_fetched_data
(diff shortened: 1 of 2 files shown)


0xgouda (approved):
Excellent!

Benchmarked using `go test -bench=. -run=^$ -count=10 -benchtime=10x` before and after, result:
```bash
benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/cybertec-postgresql/pgwatch/v6/internal/sinks
cpu: Intel(R) Core(TM) i5-10500H CPU @ 2.50GHz
                    │   old.txt   │               new.txt               │
                    │   sec/op    │   sec/op     vs base                │
WritePromMetrics-12   8.537m ± 3%   4.801m ± 5%  -43.76% (p=0.001 n=10)

                    │   old.txt    │               new.txt                │
                    │     B/op     │     B/op      vs base                │
WritePromMetrics-12   4.900Mi ± 0%   3.531Mi ± 0%  -27.95% (p=0.000 n=10)

                    │   old.txt    │               new.txt               │
                    │  allocs/op   │  allocs/op   vs base                │
WritePromMetrics-12   141.03k ± 0%   86.13k ± 0%  -38.93% (p=0.000 n=10)
```

